PR #405, opened 2018-09-24
Fixed steps for confusing apps UI in on premise SharePoint

davidchesnut:


davidchesnut:
Creating a new PR in place of this one that has a staging build

PR #424, opened 2018-09-28
Miscellaneous updates to Custom Functions articles

kbrandl:


docs/excel/custom-functions-runtime.md
```diff
@@ -1,13 +1,15 @@
...
-ms.date: 09/20/2018
+ms.date: 09/27/2018
 description: Excel custom functions use a new JavaScript runtime, which differs from the standard Add-ins WebView control runtime.
 title: Runtime for Excel custom functions
 ---
 
-# Runtime for Excel custom functions (Preview)
+# Runtime for Excel custom functions (preview)
 
 Custom functions extend Excel’s capabilities by using a new JavaScript runtime that uses a sandboxed JavaScript engine rather than a web browser. Because custom functions do not need to render UI elements, the new JavaScript runtime is optimized for performing calculations, enabling you to run thousands of custom functions simultaneously.
```
mscharlock: So I spoke with Michael on these points yesterday - he thinks: 
a) it's not important that you can run thousands simultaneously (he feels this is marketing-y)
b) the sandboxed engine in only on desktop (otherwise, web still uses WebView runtime like normal)

Maybe we say something like "On Excel's desktop client, custom functions use a new, sandboxed JavaScript engine rather than a web browser."

docs/excel/custom-functions-runtime.md
```diff
@@ -1,13 +1,15 @@
 ---
-ms.date: 09/20/2018
+ms.date: 09/27/2018
```
mscharlock: Good to know that when we update we also update the date- I will keep this in mind

docs/excel/custom-functions-overview.md
```diff
@@ -15,36 +17,133 @@ The following illustration shows an end user inserting a custom function into a
...
+// map `id` values in the JSON metadata file to the JavaScript function names
+CustomFunctionMappings.ADD = add;
+CustomFunctionMappings.INCREMENT = increment;
+```
+
+### JSON metadata file 
+
+The custom functions metadata file (**./config/customfunctions.json** in the project that the Yo Office generator creates) provides the information that Excel requires to register custom functions and make them available to end users. Custom functions are registered when a user runs an add-in for the first time. After that, they are available to that same user in all workbooks (i.e., not only in the workbook where the add-in initially ran.)
+
+> [!TIP]
```
mscharlock: This is the first time I have seen a !TIP used - are there others besides NOTE which we use frequently? I know there is a list on OPS of the options but just curious if there are some we use more frequently than others.

docs/excel/custom-functions-runtime.md
mscharlock: I will work on a PR today with some of his changes, but anyway, added this more as a heads up

PR #551, opened 2018-11-09
Add createWorkbook section

AlexJerabek:
This createWorkbook API is somewhat hidden. There's a request of the Excel team to highlight it. I'm open to suggestions for where this should go in our docs.

### DIFF
--- a/docs/excel/excel-add-ins-advanced-concepts.md
+++ b/docs/excel/excel-add-ins-advanced-concepts.md
@@ -232,6 +232,22 @@ return context.sync()
   })
 ```
 
+## Create Workbook
+
+Your add-in can create a new workbook, separate from the Excel instance in which the add-in is currently running. The Excel object has the `createWorkbook` method for this purpose.
+
+``` typescript
+createWorkbook(base64?: string): Promise<object>;
+```
+
+When this method is called, the new Workbook is immediately opened and displayed in Excel. This happens in a new Excel instance. Your add-in remains open and running with the previous workbook.
+
+``` typescript
+await Excel.createWorkbook()
+```
+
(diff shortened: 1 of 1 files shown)


docs/excel/excel-add-ins-advanced-concepts.md
```diff
@@ -232,6 +232,22 @@ return context.sync()
...
+```
+
+When this method is called, the new workbook is immediately opened and displayed in Excel. This happens in a new Excel instance. Your add-in remains open and running with the previous workbook.
+
+``` typescript
+await Excel.createWorkbook();
+```
+
+`createWorkbook` takes in an optional string. This string is an .xlsx file in base64 encoding. Assuming the string argument is a valid .xlsx file, the resulting workbook will be a copy of that file. The [FileReader](https://developer.mozilla.org/docs/Web/API/FileReader) class can be used to convert a file into the required base64 string. You can get your add-in’s current workbook as a base64 string by using [file slicing](https://docs.microsoft.com/javascript/api/office/office.document#getfileasync-filetype--options--callback-).
+
```
kbrandl: This is good info, but would be great to see examples of the scenarios you've described.  i.e., 
- example 1 > create a new/empty file
- example 2 > create a new file by passing a base64-encoded string that represents the file contents
- example 3 (optional) > create a new file that's a copy of the existing file, by using slicing to get the base64-encoded string then passing it to the `createWorkbook` function.

docs/excel/excel-add-ins-advanced-concepts.md
```diff
@@ -232,6 +232,22 @@ return context.sync()
...
+createWorkbook(base64?: string): Promise<object>;
+```
+
+When this method is called, the new workbook is immediately opened and displayed in Excel. This happens in a new Excel instance. Your add-in remains open and running with the previous workbook.
+
+``` typescript
+await Excel.createWorkbook();
+```
+
+`createWorkbook` takes in an optional string. This string is an .xlsx file in base64 encoding. Assuming the string argument is a valid .xlsx file, the resulting workbook will be a copy of that file. The [FileReader](https://developer.mozilla.org/docs/Web/API/FileReader) class can be used to convert a file into the required base64 string. You can get your add-in’s current workbook as a base64 string by using [file slicing](https://docs.microsoft.com/javascript/api/office/office.document#getfileasync-filetype--options--callback-).
```
kbrandl: change "base64 string" -> "base64-encoded string"

AlexJerabek:
Thanks for the review @kbrandl. I'll close this PR and integrate into a new Workbook article.

Issue #2238, opened 2020-09-30
Excel.CustomFunctionDimensionality.matrix is not a valid type


chucklay:
Hi,

It seems like this page is outdated. In the "Range Parameters" section, it notes that if a range is passed as a parameter, its type will be `Excel.CustomFunctionDimensionality.matrix`. While using this type, a function file written in TypeScript will fail to compile, due to `Excel` not having an exported `CustomFunctionDimensionality` member. In addition, attempting to set the type of the parameter to `any[][]` (similar to what is done in the JS example) will cause my custom function to fail due to an "Unliftable Array", which leaves me confused: is there any way to use a range as a parameter in a custom function written in typescript?

ScriptLab Gist demonstrating the missing `CustomFunctionDimensionality` (not mine, but it illustrates the issue): https://gist.github.com/maninweb/8754249beff8460abeacd6a323fddd2e

ScriptLab Gist demonstrating the "Unliftable Array" issue: https://gist.github.com/chucklay/59da1f83ae283c5685c8f6ba058cbf4a

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 17035fc0-ca69-03f2-fe37-8189ff870cc7
* Version Independent ID: 2d43e719-043b-7a62-04bb-66b2cce7bf1d
* Content: [Options for Excel custom functions - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-parameter-options?tabs=javascript#range-parameters)
* Content Source: [docs/excel/custom-functions-parameter-options.md](https://github.com/OfficeDev/office-js-docs-pr/blob/master/docs/excel/custom-functions-parameter-options.md)
* Product: **excel**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

alison-mk:
Hi @chucklay, thank you for taking the time to submit this issue! I will investigate and update the documentation, and then report back here. 

Cheers, 
Alison

chucklay:
Hi, any news on this? This has officially prevented us from publishing, as our workaround (pass the range in as a string) was ignored by Microsoft's testers, who passed the range in as a range, and failed our add-in due to the "unliftable array" issue.

alison-mk:
[...]

PR #2371, opened 2020-12-09
[Excel] (Custom Functions) Clarify Range parameters article section

alison-mk:
This PR addresses Github issue #2238.

### DIFF
--- a/docs/excel/custom-functions-parameter-options.md
+++ b/docs/excel/custom-functions-parameter-options.md
@@ -1,5 +1,5 @@
 ---
-ms.date: 11/06/2020
+ms.date: 12/09/2020
 description: 'Learn how to use different parameters within your custom functions, such as Excel ranges, optional parameters, invocation context, and more.'
 title: Options for Excel custom functions
 localization_priority: Normal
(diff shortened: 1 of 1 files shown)


docs/excel/custom-functions-parameter-options.md
```diff
@@ -114,7 +114,7 @@ function getWeatherReport(zipCode?: number, dayOfWeek?: string): string {
 
 Your custom function may accept a range of cell data as an input parameter. A function can also return a range of data. Excel will pass a range of cell data as a two-dimensional array.
 
-For example, suppose that your function returns the second highest value from a range of numbers stored in Excel. The following function accepts the parameter `values`, which is of type `Excel.CustomFunctionDimensionality.matrix`. Note that in the JSON metadata for this function, the parameter's `type` property is set to `matrix`.
+For example, suppose that your function returns the second highest value from a range of numbers stored in Excel. The following function accepts the parameter `values`, and the JSDOC syntax `number[][]` sets the parameter's `type` property to `matrix` in the JSON metadata for this function.
```
davidchesnut, line 117:
```suggestion
For example, suppose that your function returns the second highest value from a range of numbers stored in Excel. The following function accepts the parameter `values`, and the JSDOC syntax `number[][]` sets the parameter's `dimensionality` property to `matrix` in the JSON metadata for this function. 
```

The actual JSON is: "dimensionality": "matrix",
and: "type": "number"

davidchesnut (changes requested):
Nice, just one change based on observation from running in script lab.

davidchesnut (approved):
Looks great!

PR #2377, opened 2020-12-14
Add windows only note to word preview APIs

AbidRahman-MSFT:
Word preview APIs are only supported for Windows. Adding a note at the top of the page to clarify that.

### DIFF
--- a/docs/includes/word-preview-apis-note.md
+++ b/docs/includes/word-preview-apis-note.md
@@ -1,0 +1,5 @@
+> [!IMPORTANT]
+> Note that the following Word preview APIs are only available on the following platforms.
+> - Office on Windows
+>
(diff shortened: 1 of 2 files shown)


docs/includes/word-preview-apis-note.md
```diff
@@ -0,0 +1,5 @@
+> [!IMPORTANT]
+> Note that the following Word preview APIs are only available on the following platforms.
+> - Office on Windows
```
AlexJerabek, line 3: Might as well be explicit about the platform (so we don't have to worry about branding differences between Office and M365).
```suggestion
> - Word on Windows
```

AlexJerabek (approved):
Approving with one suggestion. @AbidRahman-MSFT, do you have PR merge permissions on this repo, or will you need me to check this in?

PR #2390, opened 2020-12-30
[Excel] (Custom Functions) Adjust custom functions overview article

alison-mk:
This PR: 
- Adds a shared runtime note to the CF overview article, to address [internal work item 3868818](https://office.visualstudio.com/OC/_workitems/edit/3868818/).
- Moves the Known Issues section from the Overview to the Troubleshooting article, to match the Excel information architecture: [Troubleshooting Excel Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/excel/excel-add-ins-troubleshooting#known-issues).


docs/excel/custom-functions-troubleshooting.md
```diff
@@ -72,6 +72,10 @@ CustomFunctions.associate("ADD", add);
 
 For more information on this process, see [Associating function names with JSON metadata](../excel/custom-functions-json.md#associating-function-names-with-json-metadata).
 
+## Known issues
```
AlexJerabek, line 75: There's something slightly off about having  "Common issues" precede "Known issues". The latter is using "issues" to mean bugs or deficiencies, whereas the former is referring to common design pitfalls.

docs/excel/custom-functions-troubleshooting.md
```diff
@@ -72,6 +72,10 @@ CustomFunctions.associate("ADD", add);
...
+
+See a list of additional known issues by checking the [Excel Custom Functions GitHub repository](https://github.com/OfficeDev/Excel-Custom-Functions/issues).
```
AlexJerabek, line 77: I'm not sure these are "additional" issues, rather than a complete list.
```suggestion
Known issues are tracked and reported in the [Excel Custom Functions GitHub repository](https://github.com/OfficeDev/Excel-Custom-Functions/issues).
```

docs/excel/custom-functions-overview.md
```diff
@@ -1,5 +1,5 @@
 ---
-ms.date: 11/06/2020
+ms.date: 12/29/2020
```
AlexJerabek, line 2: Since we're making changes anyway...
```suggestion
ms.date: 01/08/2020
```

docs/excel/custom-functions-troubleshooting.md
```diff
@@ -1,5 +1,5 @@
 ---
-ms.date: 11/06/2020
+ms.date: 12/29/2020
```
AlexJerabek, line 2:
```suggestion
ms.date: 01/08/2020
```

AlexJerabek (changes requested):

docs/excel/custom-functions-troubleshooting.md
alison-mk: Do you have alternate phrasing suggestions for either location? "Known issues" is oblique but it's how we title the bugs sections in add-ins docs and is in line with the industry standard, so I'm inclined to change "Common issues" instead. What do you think of "Common problems"? It avoids the confusing usage of "issues" but unfortunately is similarly vague.

AlexJerabek: Agreed that "Common issues" is the part to change. I like "Common problems", though perhaps "Common problems and solutions" sounds more enticing.

AlexJerabek (approved):
Thanks for cleaning that up. I like single place for CF concerns.

docs/excel/custom-functions-troubleshooting.md
alison-mk: Great idea!

Issue #2376, opened 2020-12-14
Clearer documentation for Enabled manifest tag platform support


brianthemike:
As the title states, I am getting an XML validation error when uploading my manifest that contains the Enabled tag. If I remove the Enabled tag, it loads fine.  Here is my XML for the control:

` <Control xsi:type="Button" id="ActionButton">`
` <Label resid="ActionButton.Label"/>`
`<Supertip>`
`<Title resid="ActionButton.Label"/>`
`<Description resid="ActionButton.Tooltip"/>`
`</Supertip>`
`<Icon>`
`<bt:Image size="16" resid="Icon.16x16"/>`
`<bt:Image size="32" resid="Icon.32x32"/>`
`<bt:Image size="80" resid="Icon.80x80"/>`
`</Icon>`
`<Action xsi:type="ExecuteFunction">`
`<FunctionName>sendSecure</FunctionName>`
`</Action>`
`<Enabled>false</Enabled>`
`</Control>`

The error being received is:
This app can't be installed. The manifest file doesn't conform to the schema definition. The element 'Control' in namespace 'http://schemas.microsoft.com/office/mailappversionoverrides/1.1' has invalid child element 'Enabled' in namespace 'http://schemas.microsoft.com/office/mailappversionoverrides/1.1'... The element 'Control' in namespace 'http://schemas.microsoft.com/office/mailappversionoverrides/1.1' has invalid child element 'Enabled' in namespace 'http://schemas.microsoft.com/office/mailappversionoverrides/1.1'.

Is this an issue with 1.1 versionoverrides?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c20c6831-2f89-5744-f1c3-5283ac6eb12e
* Version Independent ID: 400d9811-d5cc-1783-68a0-068358c37722
* Content: [Enabled element in the manifest file - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/reference/manifest/enabled)
* Content Source: [docs/reference/manifest/enabled.md](https://github.com/OfficeDev/office-js-docs-pr/blob/master/docs/reference/manifest/enabled.md)
* Product: **non-product-specific**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

AlexJerabek:
Thanks for reporting this @brianthemike. 

@exextoc, could you please investigate?

[...]

PR #2395, opened 2021-01-04
[Excel] (manifest) enabled element only valid in Excel

Rick-Kirkham:
Fixes https://github.com/OfficeDev/office-js-docs-pr/issues/2376


AlexJerabek (approved):
Looks good. I'm less familiar with the manifest documentation, but would it make sense to put that new information in a "Note" or "Important" block? It seems like an easily missed caveat, but that's from an outside perspective.

PR #2444, opened 2021-01-23
Update patterns-and-practices-overview.md

mscharlock:
Adding community call info, I feel like this would be helpful. But let me know if this feels not "docs" like enough.


davidchesnut (approved):
Nice!

PR #2458, opened 2021-01-27
[PowerPoint] (requirement set) update table in preview set

Rick-Kirkham:


AlexJerabek (approved):
Looks good. Do we need a "See also" section at the end of this article?

PR #2465, opened 2021-01-30
[Excel PowerPoint] (ribbon) new OverriddenByRibbonApi element and it's use

Rick-Kirkham:


docs/reference/manifest/overriddenbyribbonapi.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Overriding a menu item
+
+
+```xml
+<ExtensionPoint xsi:type="PrimaryCommandSurface">
+  <CustomTab id="TabCustom1">
+    <Group id="ContosoCustomTab.grp1">
+      <Control  xsi:type="Menu" id="MyMenu">
+        <!-- other child elements omitted -->
```
alison-mk, line 80:
```suggestion
        <!-- Other child elements omitted. -->
```

docs/reference/manifest/overriddenbyribbonapi.md
```diff
@@ -0,0 +1,92 @@
...
+        <Items>
+          <Item id="showGallery">
+            <OverriddenByRibbonApi>true</OverriddenByRibbonApi>
+            <!-- other child elements omitted -->
```
alison-mk, line 84:
```suggestion
            <!-- Other child elements omitted. -->
```

docs/design/contextual-tabs.md
```diff
@@ -423,9 +423,83 @@ Then your code calls the function to get the localized blob that is passed to `r
...
+
+> [!IMPORTANT]
+> Don't mark *all* of the child elements of a tab, group, or menu with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`. This is pointless if the parent element is marked with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>` for reasons given above. Moreover, if you leave out the `<OverriddenByRibbonApi>` on the parent (or set it to `false`), then the parent will appear regardless of whether custom contextual tabs are supported, but it will be empty when they are supported. So, if all the child elements shouldn't appear when custom contextual tabs are supported, mark the parent, and only the parent, with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`.
```
alison-mk, line 494: Potential option to avoid "above": 
```suggestion
> Don't mark *all* of the child elements of a tab, group, or menu with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`. This is pointless if the parent element is marked with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>` for reasons given in the preceding paragraph. Moreover, if you leave out the `<OverriddenByRibbonApi>` on the parent (or set it to `false`), then the parent will appear regardless of whether custom contextual tabs are supported, but it will be empty when they are supported. So, if all the child elements shouldn't appear when custom contextual tabs are supported, mark the parent, and only the parent, with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`.
```

alison-mk (approved):
Looks good! Approved with suggestions -- mostly copy-editing and switching "host" to "application".

PR #2466, opened 2021-02-01
[All Hosts] (extended manifest) adding conceptual docs for extended overrides

Rick-Kirkham:


docs/develop/localization.md
```diff
@@ -145,6 +145,195 @@ For Outlook add-ins, the [SourceLocation] element also aligns to the form factor
...
+     ...
+</Token>
+```
+
+But by using the wildcard character, you can have just a single `<Override>` element.
+
+```xml
+<Token xsi:type="LocaleToken" name="locale" default="en-us">
+     ...
+    <Override Locale="es-*" Value="es-es" />
```
ElizabethSamuel-MSFT, line 319: Is it possible to use just the ll e.g. `es`? If so, would the wildcard work in that instance?

docs/reference/manifest/extendedoverrides.md
```diff
@@ -7,7 +7,7 @@ localization_priority: Normal
 
 # ExtendedOverrides element
 
-Specifies the full URLs for JSON-formatted files that extend the manifest.
+Specifies the full URLs for JSON-formatted files that extend the manifest. For detailed information about the use of this element and it's descendent elements, see [Work with extended overrides of the manifest](../../develop/extended-overrides.md).
```
ElizabethSamuel-MSFT, line 10:
```suggestion
Specifies the full URLs for JSON-formatted files that extend the manifest. For detailed information about the use of this element and its descendent elements, see [Work with extended overrides of the manifest](../../develop/extended-overrides.md).
```

docs/develop/localization.md
Rick-Kirkham: No. You can't just use the ll.

docs/design/keyboard-shortcuts.md
```diff
@@ -219,6 +219,11 @@ You cannot use any of the following keyboard combinations. They are used by brow
 - Ctrl+W
```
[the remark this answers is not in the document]
Rick-Kirkham: fixed

ElizabethSamuel-MSFT (approved):

docs/develop/extended-overrides.md
```diff
@@ -0,0 +1,73 @@
...
+                "default": "CTRL+SHIFT+A"
+            }
+        }
+    ]
+}
+```
+
+## Localizing the extended overrides file
+
+If your add-in supports multiple locales, you can use the `ResourceUrl` attribute of the `<ExtendedOverrides>` element to point Office to a file of localized resources. The following is an example.
```
lindalu-MSFT, line 61:
```suggestion
If your add-in supports multiple locales, use the `ResourceUrl` attribute of the `<ExtendedOverrides>` element to point Office to a file of localized resources. The following is an example.
```
Microsoft Voice & Style guidelines: "Kill off weak writing. Most of the time, start each sentence with a verb. Cut "you can" and "there is", “there are”, “there were.” Consider reading through these topics again and remove them where _you can_. [Microsoft Voice guidelines](https://microsoft.sharepoint.com/teams/BrandCentral/Pages/The-Microsoft-brand-Identity-Voice.aspx)

docs/develop/localization.md
```diff
@@ -145,6 +145,195 @@ For Outlook add-ins, the [SourceLocation] element also aligns to the form factor
...
 
+## Localize extended overrides
+
+Some extensibility features of Office Add-ins, such as keyboard shortcuts, are configured with JSON files that are hosted on your server, instead of with the add-in's XML manifest. This section assumes that you are familiar with extended overrides. See [Work with extended overrides of the manifest](extended-overrides.md) and [ExtendedOverrides](../reference/manifest/extendedoverrides.md) element.
+
+There are two ways to localize the strings and URLs in an extended overrides file.
+
+### Method 1: Use resource tokens in the extended overrides file
+
+We recommend this method as the most efficient for most add-ins. You can use the `ResourceUrl` attribute of the [ExtendedOverrides](../reference/manifest/extendedoverrides.md) element to point Office to a file of localized resources. The following is an example.
```
lindalu-MSFT, line 156:
```suggestion
We recommend this method as the most efficient for most add-ins. Use the `ResourceUrl` attribute of the [ExtendedOverrides](../reference/manifest/extendedoverrides.md) element to point Office to a file of localized resources. The following is an example.
```

docs/develop/localization.md
```diff
@@ -145,6 +145,195 @@ For Outlook add-ins, the [SourceLocation] element also aligns to the form factor
...
+            }, 
+            "SHOWTASKPANE_action_name": {
+                "value": "Afficher le volet de tâche pour add-in",
+              } 
+        }
+    }
+}
+```
+
+Note that there is no `default` property in the file that is a peer to the `en-us` and `fr-fr` sections. This is because the default strings, which are used when the locale of the Office host application doesn't match any of the *ll-cc* properties in the resources file, *must be defined in the extended overrides file itself*. Defining the default strings directly in the extended overrides file ensures that Office doesn't have to download the resource file when the locale of the Office application matches the default locale of the add-in (as specified in the manifest). The following is a corrected version of the preceding example of an extended overrides file that uses resource tokens.
```
lindalu-MSFT, line 217:
```suggestion
There is no `default` property in the file that is a peer to the `en-us` and `fr-fr` sections. This is because the default strings, which are used when the locale of the Office host application doesn't match any of the *ll-cc* properties in the resources file, *must be defined in the extended overrides file itself*. Defining the default strings directly in the extended overrides file ensures that Office doesn't download the resource file when the locale of the Office application matches the default locale of the add-in (as specified in the manifest). The following is a corrected version of the preceding example of an extended overrides file that uses resource tokens.
```

docs/develop/localization.md
```diff
@@ -145,6 +145,195 @@ For Outlook add-ins, the [SourceLocation] element also aligns to the form factor
...
+                "value": "Show task pane for add-in",
+            } 
+        }
+    }
+}
+```
+
+### Method 2: Create separate extended override files for various locales or sets of locales
+
+If you prefer, you can create separate extended override files for each locale, rather than use a resource file. Although the [ExtendedOverrides](../reference/manifest/extendedoverrides.md) element has only one `Url` attribute, it is possible for the attribute to refer to more than one file. This is because the value of the attribute doesn't have to be a literal URL. It can be a URL template with a locale placeholder, called a token. The following is an example. Note that the format of the token is **${token.*name-of-token*}**.
```
lindalu-MSFT, line 251:
```suggestion
If you prefer, create separate extended override files for each locale, rather than use a resource file. Although the [ExtendedOverrides](../reference/manifest/extendedoverrides.md) element has only one `Url` attribute, it is possible for the attribute to refer to more than one file. This is because the value of the attribute doesn't have to be a literal URL. It can be a URL template with a locale placeholder, called a token. The following is an example. Note that the format of the token is **${token.*name-of-token*}**.
```

lindalu-MSFT (commented):
There are quite a few non-Microsoft Voice and Style issues in the text I reviewed. Please ensure all topics follow the guidelines, specifically:
1. Use of future simple tense (we use simple present)
2. Extra words, specifically "you can" everywhere
3. Mixing casual voice (contractions) with formal
4. Gerund verbs in headings
And a few others I pointed out. If you would like me to do this, please open a bug for me and I'll get to it.

docs/develop/extended-overrides.md
Rick-Kirkham: The problem with replacing "You can do x" with "Do x" is that the latter sounds like a requirement, whereas the "You can" conveys that it is an option. This sentence is a case where it is an option. (There's another way to localize with multiple locales described later in the article.)

docs/develop/localization.md
Rick-Kirkham: Here again, I want to keep the "You can" because it is not mandatory to do this. Later in the article, another way to localize is described.

docs/develop/localization.md
Rick-Kirkham: Again, this is the 2nd of two ways, so I want to keep the "You can".

docs/develop/extended-overrides.md
lindalu-MSFT: Clearly pointed out in our guidance... we are not to use "you can" if possible.  The recommendation is to use our voice to provide strong guidance. The "you can" is implied. Please remove them all to conform to the Microsoft Voice and Style.

docs/develop/localization.md
lindalu-MSFT: This is not in conformance. Either we recommend or we don't.

docs/develop/localization.md
lindalu-MSFT: Nope. It's not conformant with the guidance we are clearly given in the Microsoft Voice and Style. "You can" is considered "WEAK WRITING."

PR #2489, opened 2021-02-09
[All Hosts] (dialog) best practice with dialog.close method

Rick-Kirkham:
There will also be an update to the reference topic for Dialog.close that links to this new section.


docs/develop/dialog-best-practices.md
```diff
@@ -48,6 +48,67 @@ If, for any reason, you want to turn off this feature, then your code must opt o
...
+        else {
+         // Handle other errors
+        }
+      }
+    }
+  );
+}
+```
+
+Alternatively, you could force the code to pause before it tries to open the second dialog by using the [setTimeout](https://www.w3schools.com/jsref/met_win_settimeout.asp) method. The following is an example.
```
mscharlock, line 89:
```suggestion
// Alternatively, you could force the code to pause before it tries to open the second dialog by using the [setTimeout](https://www.w3schools.com/jsref/met_win_settimeout.asp) method. The following is an example.
```
I think you need something to indicate this is a comment

mscharlock: Actually, maybe make this regular text and separate the below into it's own chunk of code?

docs/develop/dialog-best-practices.md
```diff
@@ -48,6 +48,67 @@ If, for any reason, you want to turn off this feature, then your code must opt o
...
 
+### Opening another dialog immediately after closing one
+
+You cannot have more than one dialog open from a given host page, so your code should call [Dialog.close](/javascript/api/office/office.dialog#close__) on an open dialog before it calls `displayDialogAsync` to open another dialog. The `close` method is asynchronous. For this reason, if you call `displayDialogAsync` immediately after a call of `close`, the first dialog may not have completely closed when Office attempts to open the second. If that happens, Office will return a [12007](dialog-handle-errors-events.md#12007) error: "The operation failed because this add-in already has an active dialog."
```
mscharlock, line 53:
```suggestion
You can't have more than one dialog open from a given host page, so your code should call [Dialog.close](/javascript/api/office/office.dialog#close__) on an open dialog before it calls `displayDialogAsync` to open another dialog. The `close` method is asynchronous. For this reason, if you call `displayDialogAsync` immediately after a call of `close`, the first dialog may not have completely closed when Office attempts to open the second. If that happens, Office will return a [12007](dialog-handle-errors-events.md#12007) error: "The operation failed because this add-in already has an active dialog."
```

docs/develop/dialog-best-practices.md
```diff
@@ -48,6 +48,67 @@ If, for any reason, you want to turn off this feature, then your code must opt o
...
+
+The `close` method doesn't accept a callback parameter, and it doesn't return a Promise object so it cannot be awaited with either the `await` keyword or with a `then` method. For this reason, we suggest the following technique when you need to open a new dialog immediately after closing a dialog: encapsulate the code to open the new dialog in a method and design the method to recursively call itself if the call of `displayDialogAsync` returns `12007`. The following is a simple example.
```
mscharlock, line 55:
```suggestion
The `close` method doesn't accept a callback parameter, and it doesn't return a Promise object so it cannot be awaited with either the `await` keyword or with a `then` method. For this reason, encapsulate the code to open the new dialog in a method and design the method to recursively call itself if the call of `displayDialogAsync` returns `12007`. The following is an example.
```

mscharlock: I think I would avoid calling things simple, but that is just my preference.

mscharlock (changes requested):

docs/develop/dialog-best-practices.md
Rick-Kirkham: That's the way it is already: regular text between two code blocks/

docs/develop/dialog-best-practices.md
Rick-Kirkham: I've removed the "simple" but I kept the clause describing the scenario. This isn't going to be a common scenario and I want to make sure that people don't think they need to do this in every situation where they open a 2nd dialog. It is only needed when they open a 2nd dialog _immediately after closing a dialog_.

PR #2499, opened 2021-02-12
[All] (Hub page) Accessibility and SEO updates

alison-mk:
Accessibility and SEO updates per internal review.

### DIFF
--- a/docs/index.yml
+++ b/docs/index.yml
@@ -45,7 +45,7 @@ highlightedContent:
     # Card 7
     - itemType: reference
       title: Office JavaScript API reference documentation
-      url: https://docs.microsoft.com/javascript/api/overview
+      url: /javascript/api/overview
     # Card 8
     - itemType: sample
       title: "GitHub: Patterns and Practices for Office Add-ins"
--- a/docs/index.yml
+++ b/docs/index.yml
@@ -12,7 +12,7 @@ metadata:
   ms.topic: hub-page # Required
   author: o365devx #Required; your GitHub user alias, with correct capitalization.
   ms.author: o365devx #Required; microsoft alias of author; optional team alias.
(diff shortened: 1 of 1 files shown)


docs/index.yml
```diff
@@ -40,11 +40,11 @@ highlightedContent:
...
     # Card 6
     - itemType: how-to-guide
-      title: Beginner's guide
+      title: Office Add-ins Beginner's guide
       url: overview/learning-path-beginner.md
     # Card 7
     - itemType: reference
-      title: API reference documentation
+      title: Office JavaScript API reference documentation
       url: https://docs.microsoft.com/javascript/api/overview
```
ElizabethSamuel-MSFT, line 48: Should be able to use a relative path here
```suggestion
      url: /javascript/api/overview
```

ElizabethSamuel-MSFT (approved):
One small suggestion to use relative path for other docset

docs/index.yml
alison-mk: There are a number of docs.microsoft.com links in this file, should I make this same change for all of them? 
For example: https://docs.microsoft.com/office/dev/store/ -> /office/dev/store/

ElizabethSamuel-MSFT: Let's confirm that the link still works then update the rest

docs/index.yml
```diff
@@ -29,11 +29,11 @@ highlightedContent:
       # Card 3
     - itemType: learn
       title: "Microsoft Learn: Introduction to Office client customization with add-ins"
-      url: https://docs.microsoft.com/learn/modules/intro-office-add-ins/
```
ElizabethSamuel-MSFT, line 32: I would remove the slashes from the ends of these links

ElizabethSamuel-MSFT (approved):
just one suggestion

PR #2518, opened 2021-02-22
[Excel] (Preview APIs) Add formula changed events to intro table

alison-mk:
Update Preview APIs intro/summary table to include formula changed events.

### DIFF
--- a/docs/reference/requirement-sets/excel-preview-apis.md
+++ b/docs/reference/requirement-sets/excel-preview-apis.md
@@ -1,7 +1,7 @@
 ---
 title: Excel JavaScript preview APIs
 description: 'Details about upcoming Excel JavaScript APIs.'
-ms.date: 01/26/2021
+ms.date: 02/22/2021
 ms.prod: excel
 localization_priority: Normal
 ---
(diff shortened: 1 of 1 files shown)


docs/reference/requirement-sets/excel-preview-apis.md
```diff
@@ -16,6 +16,7 @@ The first table provides a concise summary of the APIs, while the subsequent tab
 
 | Feature area | Description | Relevant objects |
 |:--- |:--- |:--- |
+| Formula changed events | Track changes to formulas, including the source and type of event that caused a change. | [onFormulaChanged](/javascript/api/excel/excel.worksheet#onFormulaChanged)|
```
AlexJerabek, line 19: That last column typically has names of objects instead of methods. I'm okay bucking the trend if you don't want to take the suggested change, though I'd recommend prefixing the event name with `Worksheet.` in that case.
```suggestion
| Formula changed events | Track changes to formulas, including the source and type of event that caused a change. | [Worksheet](/javascript/api/excel/excel.worksheet#onFormulaChanged)|
```

AlexJerabek (approved):
Approving with a minor suggestion.

docs/reference/requirement-sets/excel-preview-apis.md
alison-mk: I will change it to Worksheet.onFormulaChanged.

Issue #2526, opened 2021-02-28
I tried to test code sample from the page.


JerryXBear:
[Enter feedback here]
I tried to test code sample from the page, but it doesn't work that  "BASICCUSTOMFUNCTION.GETRANGEVALUE threw an error: ReferenceError: 'Excel' is not defined"

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8fe047a0-5f60-f407-2ad0-7f4cc8e52082
* Version Independent ID: f2ebc9bd-25f6-cfc2-dbbe-d0090096f656
* Content: [Call Microsoft Excel APIs from a custom function - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/excel/call-excel-apis-from-custom-function)
* Content Source: [docs/excel/call-excel-apis-from-custom-function.md](https://github.com/OfficeDev/office-js-docs-pr/blob/master/docs/excel/call-excel-apis-from-custom-function.md)
* Product: **excel**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

AlexJerabek:
Thanks for reporting this @JerryXBear. 

@alison-mk, can you please take a look?

alison-mk:
Hi @JerryXBear, thank you for submitting an issue! I did some testing, and the sample on this page doesn't work with Script Lab (with Script Lab, I encountered the same ReferenceError as you). The Excel.RequestContext API described on this page only works when you have a shared JavaScript runtime set up. See [Configure your Office Add-in to use a shared JavaScript runtime](https://docs.microsoft.com/en-us/office/dev/add-ins/develop/configure-your-add-in-to-use-a-shared-runtime) to learn more. 

To avoid this confusion in the future, I've updated the [Call Microsoft Excel APIs from a custom function](https://docs.microsoft.com/en-us/office/dev/add-ins/excel/call-excel-apis-from-custom-function) article to include notes about the shared runtime requirement. This update is currently in a PR and will be live soon, so I will close this issue. 

Please let us know if you have any additional questions. 

Thanks again, 
Alison

PR #2531, opened 2021-03-04
[Excel] (Custom Functions) Update RequestContext code sample

alison-mk:
Resolves confusion reported by customer in issue #2526.

### DIFF
--- a/docs/excel/call-excel-apis-from-custom-function.md
+++ b/docs/excel/call-excel-apis-from-custom-function.md
@@ -1,35 +1,45 @@
 ---
 title: Call Microsoft Excel APIs from a custom function
 description: 'Learn which Microsoft Excel APIs you can call from your custom function.'
-ms.date: 05/11/2020
+ms.date: 03/03/2021
 localization_priority: Normal
 ---
 
 # Call Microsoft Excel APIs from a custom function
 
-Call Office.js Excel APIs from your custom functions to get range data and obtain more context for your calculations.
-
-Calling Office.js APIs through a custom function can be helpful when:
+Call Office.js Excel APIs from your custom functions to get range data and obtain more context for your calculations. Calling Office.js APIs through a custom function can be helpful when:
 
 - A custom function needs to get information from Excel before calculation. This information might include document properties, range formats, custom XML parts, a workbook name, or other Excel-specific information.
 - A custom function will set the cell's number format for the return values after calculation.
 
+> [!IMPORTANT]
+> To call Office.js Excel APIs from your custom function, you'll need to use a shared JavaScript runtime. See [Configure your Office Add-in to use a shared JavaScript runtime](../develop/configure-your-add-in-to-use-a-shared-runtime.md) to learn more.
+
 ## Code sample
 
-To call into the Office.js APIs you first need a context. Use the `Excel.RequestContext` object to get a context. Then use the context to call the APIs you need in the workbook.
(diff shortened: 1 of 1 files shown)


ElizabethSamuel-MSFT:
@alison-mk I confirmed that the build account referenced in the warning has Write access to this repo so closed and reopened PR for it to rerun build.

docs/excel/call-excel-apis-from-custom-function.md
```diff
@@ -1,35 +1,45 @@
...
+
 ## Code sample
 
-To call into the Office.js APIs you first need a context. Use the `Excel.RequestContext` object to get a context. Then use the context to call the APIs you need in the workbook.
+To call into the Office.js APIs you first need a context. Use the [Excel.RequestContext](/javascript/api/excel/excel.requestcontext) object to get a context. Then use the context to call the APIs you need in the workbook.
+
+The following code sample shows how to use `Excel.RequestContext` to access a cell through a custom function and then return the value of that cell. In this sample, the `address` is passed into the Excel JavaScript API [Worksheet.getRange](/javascript/api/excel/excel.worksheet#getRange_address_) method and must be entered as a string. For example, the custom function entered into the Excel UI must follow the pattern `=CONTOSO.GETRANGEVALUE("A1")`, where `"A1"` is the address of the cell from which you'll retrieve the value.
 
-The following code sample shows how to get a range of values from the workbook.
+> [!NOTE]
```
davidchesnut, line 24: I don't think we need this note, since this is already covered at the top of the article.

docs/excel/call-excel-apis-from-custom-function.md
```diff
@@ -1,35 +1,45 @@
...
+
 ## Code sample
 
-To call into the Office.js APIs you first need a context. Use the `Excel.RequestContext` object to get a context. Then use the context to call the APIs you need in the workbook.
+To call into the Office.js APIs you first need a context. Use the [Excel.RequestContext](/javascript/api/excel/excel.requestcontext) object to get a context. Then use the context to call the APIs you need in the workbook.
+
+The following code sample shows how to use `Excel.RequestContext` to access a cell through a custom function and then return the value of that cell. In this sample, the `address` is passed into the Excel JavaScript API [Worksheet.getRange](/javascript/api/excel/excel.worksheet#getRange_address_) method and must be entered as a string. For example, the custom function entered into the Excel UI must follow the pattern `=CONTOSO.GETRANGEVALUE("A1")`, where `"A1"` is the address of the cell from which you'll retrieve the value.
```
davidchesnut, line 22:
```suggestion
The following code sample shows how to get a value from a cell in the workbook. The address parameter is entered as a string such as `=CONTOSO.GETRANGEVALUE("A1")`, where `"A1"` is the address of the cell from which to retrieve the value.
```
A lot of this paragraph is explained by the code sample, and the code comments. I think the main point is to be sure reader knows to use a string.

davidchesnut (changes requested):
Couple comments, but looks good! Thanks!

docs/excel/call-excel-apis-from-custom-function.md
alison-mk: I see your point, and I will simplify this paragraph slightly. I found the sample very unintuitive though, so I want to err on the side of caution and include a fuller explanation in this paragraph. I also want to make sure to include the Excel JavaScript API mention + link in the paragraph, because this is why the sample requires the address parameter as a string (that getRange method takes a string).

davidchesnut (approved):
Looks great!

PR #2542, opened 2021-03-08
[Excel] (Custom Functions) Add @requiresParameterAddresses to JSON autogen article

alison-mk:


docs/excel/custom-functions-json-autogeneration.md
```diff
@@ -237,8 +235,39 @@ function getAddress(first, second, invocation) {
 }
 ```
 
-<a id="returns"></a>
+<a id="requiresParameterAddresses"></a>
+### @requiresParameterAddresses
+
+Indicates that the function should return the addresses of input parameters. 
 
+The last function parameter must be of type `CustomFunctions.Invocation` or a derived type to use  `@requiresParameterAddresses`. The JSDoc must also include an `@returns` tag specifying that the return value be a matrix, such as `@returns {string[][]}` or `@returns {number[][]}`. See [Matrix types](/office/dev/add-ins/excel/custom-functions-json-autogeneration#matrix-type) for additional information.
```
mscharlock, line 243:
```suggestion
 The JSDoc must also include an `@returns` tag specifying that the return value be a matrix, such as `@returns {string[][]}` or `@returns {number[][]}`. 
```
I'm kinda wondering if JSDoc is the right name for where the tags go?

docs/excel/custom-functions-json-autogeneration.md
```diff
@@ -320,7 +347,7 @@ A streaming function can indicate an error by calling `setResult()` with an Erro
 
 ### Promise
 
-A function can return a Promise, that provides the value when the promise is resolved. If the promise is rejected, then it will throw an error.
+A custom function can return a promise that provides the value when the promise is resolved. If the promise is rejected, then the custom function will throw an error.
```
mscharlock, line 350: Worth asking the group as a whole if we consistently use lower case p. I think that is true although I have also seen uppercase in other documentation, so I am just curious!

mscharlock (changes requested):

docs/excel/custom-functions-json-autogeneration.md
alison-mk: Ok, I'll bring it up at a team meeting! [Term studio](https://termstudio.azurewebsites.net/viewdata.aspx?tid=2254443&langCode=English) presents promise as a lowercase noun (not a proper noun).

docs/excel/custom-functions-json-autogeneration.md
alison-mk: Discussed offline -- I will adjust the JSDoc phrasing.

mscharlock (approved):

PR #2554, opened 2021-03-13
[Excel] (Workbook) Update insert worksheets article section with new Preview API 

alison-mk:
Replacing `addFromBase64` content with content for the new `insertWorksheetsFromBase64` preview API.

### DIFF
--- a/docs/excel/excel-add-ins-workbooks.md
+++ b/docs/excel/excel-add-ins-workbooks.md
@@ -82,7 +82,7 @@ insertWorksheetsFromBase64(base64File: string, options?: Excel.InsertWorksheetOp
 The following example inserts another workbook into the current workbook. The new worksheets are inserted after the active worksheet. Note that `[]` is passed as the parameter for the [InsertWorksheetOptions](/javascript/api/excel/excel.insertworksheetoptions) `sheetNamesToInsert` property. This means that all the worksheets from the existing workbook are inserted into the current workbook.
 
 > [!IMPORTANT]
-> The `insertWorksheetsFromBase64` method is supported for Excel on Windows, Mac, and the web. It's not supported for iOS. Additionally, in Excel on the web this method doesn't support source worksheets with PivotTable, Chart, Comment, or Slicer elements. If those objects are present, the `insertWorksheetsFromBase64` method Excel on the web returns the `UnsupportedFeature` error. 
+> The `insertWorksheetsFromBase64` method is supported for Excel on Windows, Mac, and the web. It's not supported for iOS. Additionally, in Excel on the web this method doesn't support source worksheets with PivotTable, Chart, Comment, or Slicer elements. If those objects are present, the `insertWorksheetsFromBase64` method returns the `UnsupportedFeature` error in Excel on the web. 
 
 ```js
 var myFile = document.getElementById("file");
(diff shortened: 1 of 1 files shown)


docs/excel/excel-add-ins-workbooks.md
```diff
@@ -70,34 +70,42 @@ reader.readAsDataURL(myFile.files[0]);
...
 
 ```TypeScript
-addFromBase64(base64File: string, sheetNamesToInsert?: string[], positionType?: Excel.WorksheetPositionType, relativeTo?: Worksheet | string): OfficeExtension.ClientResult<string[]>;
+insertWorksheetsFromBase64(base64File: string, options?: Excel.InsertWorksheetOptions): OfficeExtension.ClientResult<string[]>;
 ```
 
-The following example shows a workbook's worksheets being inserted in the current workbook, directly after the active worksheet. Note that `null` is passed for the `sheetNamesToInsert?: string[]` parameter. This means all the worksheets are being inserted.
+The following example shows an existing workbook's worksheets being inserted in the current workbook, directly after the active worksheet. Note that `[]` is passed as the parameter for the [InsertWorksheetOptions](/javascript/api/excel/excel.insertworksheetoptions) `sheetNamesToInsert` property. This means that all the worksheets from the existing workbook should be inserted into the current workbook.
```
AlexJerabek, line 82: Suggestion to remove the possessive (which helps localization). Also changing the "should" to something more definitive.
```suggestion
The following example inserts another workbook in the current workbook. The new worksheets are inserted after the active worksheet. Note that `[]` is passed as the parameter for the [InsertWorksheetOptions](/javascript/api/excel/excel.insertworksheetoptions) `sheetNamesToInsert` property. This means that all the worksheets from the existing workbook are inserted into the current workbook.
```

docs/excel/excel-add-ins-workbooks.md
```diff
@@ -70,34 +70,42 @@ reader.readAsDataURL(myFile.files[0]);
...
+
+> [!IMPORTANT]
+> The `insertWorksheetsFromBase64` method is supported for Excel on Windows, Mac, and the web. It's not supported for iOS. In addition, in Excel on the web this method doesn't support source worksheets with PivotTable, Chart, Comment, or Slicer elements. Using the `insertWorksheetsFromBase64` method to insert worksheets with these unsupported features in Excel on the web will return the `UnsupportedFeature` error.
```
AlexJerabek, line 85: Suggestions for brevity and removing the "-ing" verb (as per the style guide).
```suggestion
> The `insertWorksheetsFromBase64` method is supported for Excel on Windows, Mac, and the web. It's not supported for iOS. Additionally, in Excel on the web this method doesn't support source worksheets with PivotTable, Chart, Comment, or Slicer elements. If those objects are present, the `insertWorksheetsFromBase64` method Excel on the web returns the `UnsupportedFeature` error. 
```

AlexJerabek (approved):
Looks great! I left a couple minor phrasing suggestions, but this should be good to go otherwise.

docs/excel/excel-add-ins-workbooks.md
alison-mk: Great suggestions, thanks Alex!

PR #2595, opened 2021-03-31
Update container folder name to reflect latest Office changes.

jargil:
Office's containers are now named after the product and not the identifier.


jargil:
Abandoning this PR.

PR #2602, opened 2021-04-06
[Admin] Include en in Wikipedia links

ElizabethSamuel-MSFT:
Due to nature of Wikipedia, locale-agnostic links don't work as might be expected.


Rick-Kirkham (approved):
Approving with a question: 
Are we leaving the metadata the dates unchanged because this is so trivial?

ElizabethSamuel-MSFT:
> Approving with a question:
> Are we leaving the metadata the dates unchanged because this is so trivial?

My approach has been to update the ms.date if there's a visible change. The underlying link is not visible so I didn't update the ms.date.

PR #2603, opened 2021-04-06
[Shared runtime] Requirement set not supported on Outlook

ElizabethSamuel-MSFT:


docs/develop/configure-your-add-in-to-use-a-shared-runtime.md
```diff
@@ -218,7 +219,7 @@ Don't design your add-in to use multiple task panes if you are planning to use a
 
 ## Give us feedback
 
-We'd love to hear your feedback on this feature. If you find any bugs, issues, or have requests on this feature, please let us know by creating a GitHub issue in the [office-js repo](https://github.com/OfficeDev/office-js).
+We'd love to hear your feedback on this feature. If you find any bugs, issues, or have requests on this feature, please let us know by going to the **Feedback** section at the bottom of the page, then selecting **This page** to create a GitHub issue.
```
davidchesnut, line 222: This should continue to point to the office-js repo because that's where we gather feedback on preview features. the Runtime team is not checking docs for feature feedback.

docs/includes/shared-runtime-requirements-note.md
```diff
@@ -1,6 +1,9 @@
 > [!IMPORTANT]
-> Note that the shared JavaScript runtime is only available on the following platforms.
+> Note that the shared JavaScript runtime requirement set is only available on the following platforms.
+>
 > - Excel on the web, Windows, and Mac.
 > - PowerPoint on Windows (build 13218.10000 or later). The shared JavaScript runtime for PowerPoint is currently in preview and subject to change. It is not supported for use in production environments. To get the latest build you need to [join Office Insider](https://insider.office.com/join). A good way to try out preview features is by using a Microsoft 365 subscription. If you don't already have a Microsoft 365 subscription, you can get one by joining the [Microsoft 365 developer program](https://developer.microsoft.com/office/dev-program).
 >
 > At this time, the shared JavaScript runtime is not supported on iPad or in one-time purchase versions of Office 2019 or earlier.
+>
```
davidchesnut, line 8: Would be preferable to not bring in the event-based activation discussion into the note since the note shows up in articles that don't have anything to do with Outlook. Would be nice if we could keep it to a once sentence "Outlook doesn't use the SharedRuntime requirement set. For more info see...." approach.

docs/reference/manifest/runtime.md
```diff
@@ -1,22 +1,17 @@
 ---
 title: Runtime in the manifest file
 description: The Runtime element configures your add-in to use a shared JavaScript runtime for its various components, for example, ribbon, task pane, custom functions.
-ms.date: 05/29/2020
+ms.date: 04/07/2021
 localization_priority: Normal
 ---
 
-# Runtime element (preview)
+# Runtime element
```
davidchesnut, line 8: I think I'd modify this to talk about how the Runtime is used for "shared runtime" and also for "event-based activation". These are the two separate things which definitely apply to Runtime and make more sense then "here's how to use it in Excel, and here's how to use it in Outlook". example:

"For the shared runtime, this element enables the ribbon, task pane, and custom functions to use the same runtime..."
"For event based activation, this element enables your add-in to receive events from Outlook such as a new message created event..."

Then the include note about which Office apps support the shared runtime.

docs/reference/manifest/runtime.md
```diff
@@ -40,3 +35,5 @@ In Outlook, this element enables event-based add-in activation. For more informa
 ## See also
 
 - [Runtimes](runtimes.md)
+- [Configure your Office Add-in to use a shared JavaScript runtime](../../develop/configure-your-add-in-to-use-a-shared-runtime.md)
```
davidchesnut, line 38: Nice!

davidchesnut (changes requested):

docs/reference/manifest/runtime.md
ElizabethSamuel-MSFT: I added a different include note for the manifest elements

docs/reference/manifest/runtime.md
ElizabethSamuel-MSFT: Thx :)

docs/reference/manifest/runtime.md
davidchesnut: I like this!

davidchesnut (approved):

PR #2616, opened 2021-04-12
Update from Adam/Igor about CF UI-less

mscharlock:
@alison-mk we should check that this is only for desktop and not available online as well


alison-mk (approved):

AlexJerabek:
@mscharlock Can this PR be merged?

Issue #2624, opened 2021-04-16
Runtime Environment Informations


heebinho:
That first Note "When running in Office on Windows, your add-in uses the Internet Explorer 11 browser." means, that as soon as you have specified it, an Add-In will use IE11 **regardless** of the Office version documented in [Browsers used by Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/concepts/browsers-used-by-office-web-add-ins), right?

So in my case IE11 is used instead of Microsoft Edge with original WebView (EdgeHTML).

How does that look in the future with the chromium based Edge or is this Information already outdated?

Any additional informations regarding this topic would be highly appreciated.
Some design decisions depend on the embedded browser controls. 

Thank you in advance.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4a8cf67b-951d-7bae-5584-565ae9ec45c1
* Version Independent ID: b7b60c57-bcb0-679e-5163-97b62fced5d3
* Content: [Runtimes in the manifest file - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/reference/manifest/runtimes)
* Content Source: [docs/reference/manifest/runtimes.md](https://github.com/OfficeDev/office-js-docs-pr/blob/master/docs/reference/manifest/runtimes.md)
* Product: **non-product-specific**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

AlexJerabek:
Thanks for reporting this @heebinho. 

@Rick-Kirkham, can you please take a look at the [Runtimes manifest article](https://docs.microsoft.com/en-us/office/dev/add-ins/reference/manifest/runtimes)? That note does seem out of date.

Rick-Kirkham:
@heebinho  You are correct. The note is out-of-date. A Shared Runtime will now use Edge (Chromium) on the same Windows and Office version combinations that would support Edge (Chromium) in a task pane. I will get this article updated.

PR #2628, opened 2021-04-17
[All Hosts] (manifest) update Runtimes element

Rick-Kirkham:
Fixes https://github.com/OfficeDev/office-js-docs-pr/issues/2624


docs/reference/manifest/runtimes.md
```diff
@@ -10,7 +11,7 @@ localization_priority: Normal
 Specifies the runtime of your add-in. Child of the [`<Host>`](host.md) element.
 
 > [!NOTE]
-> When running in Office on Windows, your add-in uses the Internet Explorer 11 browser.
+> When running in Office on Windows, an add-in that has a `<Runtimes>` element in its manifest does not necessarily run in the same webview control as it otherwise would. The article [Browsers used by Office Add-ins](../../concepts/browsers-used-by-office-web-add-ins.md) describes how the versions of Windows and Office determine what webview control is normally used. If the conditions described there for using Microsoft Edge with WebView2 (Chromium-based) are met, then the add-in uses that browser whether or not it has a `<Runtimes>` element. However, when those conditions are not met, an add-in with a `<Runtimes>` element always uses Internet Explorer 11 regardless of the Windows or Microsoft 365 version.
```
lindalu-MSFT, line 14:
```suggestion
> When running in Office on Windows, an add-in that has a `<Runtimes>` element in its manifest does not necessarily run in the same webview control as it otherwise would. For more information about how the versions of Windows and Office determine what webview control is normally used, see [Browsers used by Office Add-ins](../../concepts/browsers-used-by-office-web-add-ins.md). If the conditions described there for using Microsoft Edge with WebView2 (Chromium-based) are met, then the add-in uses that browser whether or not it has a `<Runtimes>` element. However, when those conditions are not met, an add-in with a `<Runtimes>` element always uses Internet Explorer 11 regardless of the Windows or Microsoft 365 version.
```
For accessibility, describe the link content before the link. That's why we use the "For more information about blah, see [blah].

lindalu-MSFT (approved):

PR #2636, opened 2021-04-22
[All hosts](Sideloading) Add note for which manifest to use from Visual Studio

davidchesnut:
Added a note to clarify which manifest to use when sideloading from Visual Studio 2019 projects.


mscharlock:
I'm going to assume this information is correct - I haven't verified it but the note seems fine!

mscharlock (approved):

PR #2651, opened 2021-04-30
[Excel] (add-in commands) Remove misleading awaits

Rick-Kirkham:


docs/design/disable-add-in-commands.md
```diff
@@ -107,11 +107,11 @@ const enableButton = async () => {
     const parentGroup: Group = {id: "CustomGroup111", controls: [button]};
     const parentTab: Tab = {id: "OfficeAddinTab1", groups: [parentGroup]};
     const ribbonUpdater: RibbonUpdaterData = { tabs: [parentTab]};
-    await Office.ribbon.requestUpdate(ribbonUpdater);
+    Office.ribbon.requestUpdate(ribbonUpdater);
 }
 ```
 
-Office controls when it updates the state of the ribbon. The **requestUpdate()** method queues a request to update. The method will resolve the Promise object as soon as it has queued the request, not when the ribbon actually updates.
+You can `await` the call of **requestUpdate()** if the parent function is asynchronous, but note that Office controls when it updates the state of the ribbon. The **requestUpdate()** method queues a request to update. The method will resolve the promise object as soon as it has queued the request, not when the ribbon actually updates.
```
ElizabethSamuel-MSFT, line 114:
```suggestion
You can `await` the call of **requestUpdate()** if the parent function is asynchronous, but note that the Office application controls when it updates the state of the ribbon. The **requestUpdate()** method queues a request to update. The method will resolve the promise object as soon as it has queued the request, not when the ribbon actually updates.
```

ElizabethSamuel-MSFT (approved):
One small comment...

PR #2673, opened 2021-05-11
[Outlook] Requirement set 1.10

ElizabethSamuel-MSFT:


docs/toc.yml
```diff
@@ -865,6 +865,20 @@ items:
           href: reference/objectmodel/preview-requirement-set/Office.context.mailbox.item.md
         - name: Outlook Preview API Reference
           href: reference/outlook-api-reference-preview.md
+      - name: Requirement Set 1.10
```
lindalu-MSFT, line 868: Super nit-picky thing, but this really should be "Requirement set" in the TOC! All headings and titles should be sentence case and "requirement set" is lower case. If you don't feel like changing them all (I don't blame you) then consider changing the top level TOC to match casing.
https://review.docs.microsoft.com/en-us/office/dev/add-ins/reference/objectmodel/requirement-set-1.10/outlook-requirement-set-1.10?branch=outlook-mailbox-1.10

lindalu-MSFT (approved):

PR #2678, opened 2021-05-13
Update resources in index.yml

lindalu-MSFT:
Change SO to Q&A
Change UV to TC Ideas forum

### DIFF
--- a/docs/outlook/index.yml
+++ b/docs/outlook/index.yml
@@ -60,7 +60,7 @@ landingContent:
       - linkListType: reference
         links:
           - text: Ask questions
-            url: https://docs.microsoft.com/en-us/answers/topics/office-addins-dev.html
+            url: /answers/topics/office-addins-dev.html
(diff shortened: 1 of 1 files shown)


docs/outlook/index.yml
```diff
@@ -60,9 +60,9 @@ landingContent:
       - linkListType: reference
         links:
           - text: Ask questions
-            url: https://stackoverflow.com/questions/tagged/office-js
+            url: https://docs.microsoft.com/en-us/answers/topics/office-addins-dev.html
```
AlexJerabek, line 63: You can link to docs.microsoft.com content like this. It allows for the link to be accessible in air-gapped environments.
```suggestion
            url: /answers/topics/office-addins-dev.html
```

AlexJerabek (approved):
Thanks @lindalu-MSFT. Please change the link as requested, but I don't need to re-review it afterwards.

Issue #2657, opened 2021-05-04
Event based activation - Are listed events correct / enabled?


barclayadam:
I'm trying to get some event based add-in working but cannot seem to get anything other than `OnNewMessageCompose` working for email-based events. Neither `OnMessageAttachmentsChanged` nor `OnMessageRecipientsChanged` ever seem to be fired, but based on the table at the top of this document the events seem to be supported.

To eliminate possible problems with my implementation I have a simple HTML file with a simple function to print to console, nothing else. In my manifest.xml I have:

````xml
            <Hosts>
                <Host xsi:type="MailHost">
                    <!-- Event-based activation happens in a lightweight runtime.-->
                    <Runtimes>
                        <!-- HTML file including reference to or inline JavaScript event handlers.
                             This is used by Outlook on the web. -->
                        <Runtime resid="WebViewRuntime.Url">
                            <!-- JavaScript file containing event handlers. This is used by Outlook Desktop. -->
                            <Override type="javascript" resid="JSRuntime.Url"/>
                        </Runtime>
                    </Runtimes>

                    <DesktopFormFactor>
                        <!-- Enable launching the add-in on the included events. -->
                        <ExtensionPoint xsi:type="LaunchEvent">
                            <LaunchEvents>
                                <LaunchEvent Type="OnNewMessageCompose" FunctionName="printToConsole" />
                                <LaunchEvent Type="OnMessageRecipientsChanged" FunctionName="printToConsole" />
                            </LaunchEvents>
                            <!-- Identifies the runtime to be used (also referenced by the Runtime element). -->
                            <SourceLocation resid="WebViewRuntime.Url"/>
                        </ExtensionPoint>
                    </DesktopFormFactor>
                </Host>
            </Hosts>
````

When installing to a preview-enabled OWA installation the function *is* called when a new message is composed, but *is NOT* called when recipients or attachments change.

That seems to be correct based on the document at https://docs.microsoft.com/en-us/office/dev/add-ins/reference/manifest/extensionpoint#launchevent-preview which suggests the only value LaunchEvent types are `OnNewMessageCompose` and `OnNewAppointmentOrganizer`

To be clear on my use case I wish to activate my add-in when a new email is composed, and then continue to react to events such as recipients or subject changing and run some additional code. I have tried to add event handlers directly within the body of the `OnNewMessageCompose` function but that, although claiming to be successful, never actually fires.

[...]

PR #2680, opened 2021-05-13
[Outlook] (preview) Update event-based activation walkthrough

ElizabethSamuel-MSFT:
Related to issue https://github.com/OfficeDev/office-js-docs-pr/issues/2657.


docs/outlook/autolaunch.md
```diff
@@ -46,7 +46,9 @@ To preview this feature:
   - The minimum required build is 16.0.14026.20000. Join the [Office Insider program](https://insider.office.com) for access to Office beta builds.
   - Configure the registry:
```
JuaneloJuanelo, line 47: lets explain why this is needed.  "In Outlook for Win32 we download a copy of Office.js (as opposed to load it from the CDN) . By default we download the Production version of the API. If you want to use the BETA version of the  Outlook JavaScript APIs that are in Beta you need to add this registry entry, otherwise the API will not be found. "

PR #2687, opened 2021-05-14
[Outlook] (preview) Runtime Override element

ElizabethSamuel-MSFT:
Related to issue https://github.com/OfficeDev/office-js-docs-pr/issues/2657

### DIFF
--- a/docs/reference/manifest/override.md
+++ b/docs/reference/manifest/override.md
@@ -180,7 +180,7 @@ An `<Override>` element expresses a conditional and can be read as an "If ... th
 </ExtendedOverrides>
 ```
 
-## See also
+### See also
 
 - [Office versions and requirement sets](../../develop/office-versions-and-requirement-sets.md)
 - [Set the Requirements element in the manifest](../../develop/specify-office-hosts-and-api-requirements.md#set-the-requirements-element-in-the-manifest)
--- a/docs/reference/manifest/override.md
+++ b/docs/reference/manifest/override.md
@@ -189,7 +189,7 @@ An `<Override>` element expresses a conditional and can be read as an "If ... th
 ## Override element of type RuntimeOverride (preview)
 
 > [!IMPORTANT]
-> This feature is only supported for [preview](../reference/objectmodel/preview-requirement-set/outlook-requirement-set-preview.md) in Outlook on the web and on Windows with a Microsoft 365 subscription. For more details, see [Configure your Outlook add-in for event-based activation](../../outlook/autolaunch.md).
+> This feature is only supported for [preview](../../reference/objectmodel/preview-requirement-set/outlook-requirement-set-preview.md) in Outlook on the web and on Windows with a Microsoft 365 subscription. For more details, see [Configure your Outlook add-in for event-based activation](../../outlook/autolaunch.md).
 >
 > Because preview features are subject to change without notice, they shouldn't be used in production add-ins.
 
--- a/docs/reference/manifest/override.md
+++ b/docs/reference/manifest/override.md
@@ -15,7 +15,7 @@ Provides a way to override the value of a manifest setting depending on a specif
 
 An `<Override>` element that is inside of a `<Runtime>` element must be of type **RuntimeOverride**.
 
-There is no `type` parameter for the other types of `<Override>` element. The difference is determined by the parent element and the parent element's type. An `<Override>` element that is inside of a `<Token>` element whose `xsi:type` is `RequirementToken`, must be of type **RequirementTokenOverride**. An `<Override>` element inside any other parent element, or inside an `<Override>` element of type `LocaleToken`, must be of type **LocaleTokenOverride**. For more information about the use of this element when it is a child of a `<Token>` element, see [Work with extended overrides of the manifest](../../develop/extended-overrides.md).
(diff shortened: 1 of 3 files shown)


docs/reference/manifest/override.md
```diff
@@ -1,18 +1,19 @@
...
+There are three types of `<Override>` elements, one is for locale overrides, called **LocaleTokenOverride**, another for requirement set overrides, called **RequirementTokenOverride**, and a third for runtime overrides, called **RuntimeOverride**. But there is no `type` parameter for the first two types of `<Override>` element. The difference is determined by the parent element and the parent element's type. An `<Override>` element that is inside of a `<Token>` element whose `xsi:type` is `RequirementToken`, must be of type **RequirementTokenOverride**. An `<Override>` element that is inside of a `<Runtime>` element must be of type **RuntimeOverride**. An `<Override>` element inside any other parent element, or inside an `<Override>` element of type `LocaleToken`, must be of type **LocaleTokenOverride**. Each type is described in separate sections later in this article. For more information about the use of this element when it is a child of a `<Token>` element, see [Work with extended overrides of the manifest](../../develop/extended-overrides.md).
```
davidchesnut, line 16: This first sentence is repeating what the bullet list says. I'd combine the names up into the bullet list and get rid of this sentence. So like the first bullet item would read "An Office local that is different from the default, called **LocalTokenOverride**"...

docs/reference/manifest/override.md
```diff
@@ -184,3 +185,56 @@ An `<Override>` element expresses a conditional and can be read as an "If ... th
...
+### Contained in
+
+- [Runtime](runtime.md)
+
+### Attributes
+
+|Attribute|Type|Required|Description|
+|:-----|:-----|:-----|:-----|
+|**type**|string|Yes|Specifies the language for this override. At present, `"javascript"` is the only supported option.|
+|**resid**|string|Yes|Specifies the URL location of the JavaScript file that should override the URL location of the default HTML defined in the parent `Runtime` element's `resid`. The `resid` can be no more than 32 characters and must match an `id` attribute of a `Url` element in the `Resources` element.|
```
davidchesnut, line 221: Can we make the elements link to their pages when possible? Like `Runtime` should link to the Runtime element page. Makes it easier to follow the dots when working on the manifest.

docs/reference/manifest/override.md
```diff
@@ -184,3 +185,56 @@ An `<Override>` element expresses a conditional and can be read as an "If ... th
...
 - [Keyboard shortcuts](../../design/keyboard-shortcuts.md)
+
+## Override element of type RuntimeOverride (preview)
+
+> [!IMPORTANT]
+> This feature is only supported for [preview](../../reference/objectmodel/preview-requirement-set/outlook-requirement-set-preview.md) in Outlook on the web and on Windows with a Microsoft 365 subscription. For more details, see [Configure your Outlook add-in for event-based activation](../../outlook/autolaunch.md).
+>
+> Because preview features are subject to change without notice, they shouldn't be used in production add-ins.
+
+An `<Override>` element expresses a conditional and can be read as an "If ... then ..." statement. If the `<Override>` element is of type **RuntimeOverride**, then the `type` attribute is the condition, and the `resid` attribute is the consequent. For example, the following is read "If the type is 'javascript', then the `resid` is 'JSRuntime.Url'." Outlook Desktop requires this element for [LaunchEvent extension point](../../reference/manifest/extensionpoint.md#launchevent-preview) handlers.
```
davidchesnut, line 196: I'm not following what "RuntimeOverride" is. I don't see it in the syntax or example. Is this an element? I notice the other two "types" also have names like this. LocaleTokenOverride for example. If these are real types, we should link to the pages for them. If these are just conceptual names for how the Override element is used, we should just refer to them like "Override for Runtime element" and "Override for locale token" and so on.

davidchesnut (changes requested):

docs/reference/manifest/override.md
Rick-Kirkham: So, now there IS a `type` attribute, but it is used to distinguish runtime types, not override types, so we need to reword this. 

```suggestion
There are three types of `<Override>` elements, one is for locale overrides, called **LocaleTokenOverride**, another for requirement set overrides, called **RequirementTokenOverride**, and a third for runtime overrides, called **RuntimeOverride**. But there is no `overrideType` attribute for the `<Override>` element. The difference is determined by the parent element and the parent element's type. An `<Override>` element that is inside of a `<Token>` element whose `xsi:type` is `RequirementToken`, must be of type **RequirementTokenOverride**. An `<Override>` element that is inside of a `<Runtime>` element must be of type **RuntimeOverride**. An `<Override>` element inside any other parent element, or inside an `<Override>` element of type `LocaleToken`, must be of type **LocaleTokenOverride**. Each type is described in separate sections later in this article. For more information about the use of this element when it is a child of a `<Token>` element, see [Work with extended overrides of the manifest](../../develop/extended-overrides.md).
```

docs/reference/manifest/override.md
Rick-Kirkham: > I'm not following what "RuntimeOverride" is. I don't see it in the syntax or example. Is this an element? I notice the other two "types" also have names like this. LocaleTokenOverride for example. If these are real types, we should link to the pages for them. If these are just conceptual names for how the Override element is used, we should just refer to them like "Override for Runtime element" and "Override for locale token" and so on.

RuntimeOverride is a type in the official XSD. (go to [MS-OWEMXML](https://docs.microsoft.com/en-us/openspecs/office_file_formats/ms-owemxml/82e93ec5-de22-42a8-86e3-353c8336aa40) and search for "RuntimeOverride". 

The other two are notional "types" that are not formally defined. They are implicitly defined by what their parent element is.  This is inconsistent and confusing. I think we should push back on Zlatko about this design.

Rick-Kirkham (changes requested):

Rick-Kirkham (approved):

davidchesnut (approved):
Looks great!

PR #2715, opened 2021-05-28
[Excel] (Workbook) Adjust insert worksheets article section

alison-mk:
This update incorporates internal feedback, more closely aligns the insertWorksheetsFromBase64 conceptual docs with the Script Lab code sample, and updates a related code sample in the article for consistency.


docs/excel/excel-add-ins-workbooks.md
```diff
@@ -80,37 +81,39 @@ The previous example shows a new workbook being created from an existing workboo
...
+// Retrieve the external workbook file and set up a `FileReader` object. 
 var myFile = document.getElementById("file");
 var reader = new FileReader();
 
 reader.onload = (event) => {
     Excel.run((context) => {
         // Remove the metadata before the base64-encoded string.
         const startIndex = reader.result.toString().indexOf("base64,");
-        const workbookContents = reader.result.toString().substr(startIndex + 7);
+        const externalWorkbook = reader.result.toString().substr(startIndex + 7);
```
AlexJerabek, line 98: We have an outstanding work item to "modernize" some of the JS samples to use `const` and `let` instead of `var`, but this is the only sample doing that. For now, let's make these variable declarations `var` and update the Excel article code in a different, single effort.

docs/excel/excel-add-ins-workbooks.md
```diff
@@ -80,37 +81,39 @@ The previous example shows a new workbook being created from an existing workboo
...
-
 > [!IMPORTANT]
 > The `insertWorksheetsFromBase64` method is supported for Excel on Windows, Mac, and the web. It's not supported for iOS. Additionally, in Excel on the web this method doesn't support source worksheets with PivotTable, Chart, Comment, or Slicer elements. If those objects are present, the `insertWorksheetsFromBase64` method returns the `UnsupportedFeature` error in Excel on the web. 
 
+The following code sample builds on the preceding code sample and shows how to insert worksheets from another workbook into the current workbook. The following code sample first processes a workbook file with a [`FileReader`](https://developer.mozilla.org/docs/Web/API/FileReader) object and extracts a base64-encoded string, and then it inserts this base64-encoded string into the current workbook. The new worksheets are inserted after the worksheet named **Sheet1**. Note that `[]` is passed as the parameter for the [InsertWorksheetOptions](/javascript/api/excel/excel.insertworksheetoptions) `sheetNamesToInsert` property. This means that all the worksheets from the target workbook are inserted into the current workbook.
```
AlexJerabek, line 87: I'm not sure this sample builds on the previous one as much as it provides an alternate solution. Consider rephrasing how you connect the to samples (if at all).

AlexJerabek, line 87: This was an issue in the original text, but there's something about "[InsertWorksheetOptions](/javascript/api/excel/excel.insertworksheetoptions) `sheetNamesToInsert` property" that seems off. It's that the property isn't attached with a dot. How about "[InsertWorksheetOptions.sheetNamesToInsert](/javascript/api/excel/excel.insertworksheetoptions) property"

AlexJerabek (changes requested):
I like the sample, just curious how we can improve the links between this and the other sample.

AlexJerabek (approved):

PR #2716, opened 2021-05-28
replace log in with sign in

lindalu-MSFT:
Per https://office.visualstudio.com/OC/_workitems/edit/4398231?src=WorkItemMention&src-action=artifact_link


docs/develop/overview-authn-authz.md
```diff
@@ -14,7 +14,7 @@ Your add-in can also get the user's consent to access their Microsoft Graph data
...
+- **Web Application Authentication and Authorization with Azure Active Directory**: This isn't something new or special. It's just the way Office Add-ins (and other web apps) authenticated users and authorized apps before there was an Office SSO system, and it is still used in scenarios where Office SSO cannot be. Also, there are scenarios in which you want to have your users sign in to your add-in separately even when SSO is available; for example, if you want them to have the option of signing in to the add-in with a different ID from the one with which they are currently logged in to Office.
```
Rick-Kirkham, line 17:
```suggestion
- **Web Application Authentication and Authorization with Azure Active Directory**: This isn't something new or special. It's just the way Office Add-ins (and other web apps) authenticated users and authorized apps before there was an Office SSO system, and it is still used in scenarios where Office SSO cannot be. Also, there are scenarios in which you want to have your users sign in to your add-in separately even when SSO is available; for example, if you want them to have the option of signing in to the add-in with a different ID from the one with which they are currently signed in to Office.
```

Rick-Kirkham (approved):
Approving with one suggestion.

davidchesnut (approved):

PR #2732, opened 2021-06-05
[Excel] (Workbook) Add onActivated conceptual docs

alison-mk:
This PR adds:
- A note about Workbook.onActivated in the Events article
- An article section about onActivated in the Workbook article

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,6 @@ _site/
 .optemp/
 _themes*/
 _repo.*/
+*.DS_Store
 
 .openpublishing.buildcore.ps1
--- a/docs/excel/excel-add-ins-workbooks.md
+++ b/docs/excel/excel-add-ins-workbooks.md
@@ -341,9 +341,17 @@ context.application.suspendApiCalculationUntilNextSync();
 ## Detect workbook activation (preview)
 
 > [!NOTE]
-> The `Worksheet.onActivated` event is currently only available in public preview. [!INCLUDE [Information about using preview APIs](../includes/using-excel-preview-apis.md)]
+> The `Workbook.onActivated` event is currently only available in public preview. [!INCLUDE [Information about using preview APIs](../includes/using-excel-preview-apis.md)]
 > 
 
+Your add-in can detect when a workbook is activated. After a workbook is opened, it will become "inactive" if the user switches focus to another workbook, to another application, or (in Excel on the web) to another tab of the web browser. The workbook is "activated" when the user switches focus back to this workbook.
+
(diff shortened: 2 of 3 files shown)


docs/excel/excel-add-ins-workbooks.md
```diff
@@ -338,6 +338,40 @@ The Excel API also lets add-ins turn off calculations until `RequestContext.sync
 context.application.suspendApiCalculationUntilNextSync();
 ```
 
+## Detect workbook activation (preview)
+
+> [!NOTE]
+> The `Workbook.onActivated` event is currently only available in public preview. [!INCLUDE [Information about using preview APIs](../includes/using-excel-preview-apis.md)]
+> 
+
+Your add-in can detect when a workbook is activated. After a workbook is opened, it will become "inactive" if the user switches focus to another workbook, to another application, or (in Excel on the web) to another tab of the web browser. A workbook is "activated" when the user switches focus back to this workbook. After detecting workbook activation, your add-in can trigger callback functions, such as refreshing workbook data.
```
AlexJerabek, line 347: Some phrasing suggestions. Also, I think that bold or italics are better for emphasis, rather than quotes.
```suggestion
Your add-in can detect when a workbook is activated. A workbook becomes **inactive** when the user switches focus to another workbook, to another application, or (in Excel on the web) to another tab of the web browser. A workbook is **activated** when the user returns focus to the workbook. The workbook activation can trigger callback functions your add-in, such as refreshing workbook data.
```

.gitignore
```diff
@@ -4,5 +4,6 @@ _site/
 .optemp/
 _themes*/
 _repo.*/
+*.DS_Store
```
AlexJerabek, line 7: Was this intentional?

docs/excel/excel-add-ins-workbooks.md
```diff
@@ -338,6 +338,40 @@ The Excel API also lets add-ins turn off calculations until `RequestContext.sync
...
+    // Register the workbook activated event handler.
+    workbook.onActivated.add(workbookActivated);
+
+    return context.sync();
+});
+
+function workbookActivated(event) {
+  Excel.run(function (context) {
+    // Callback function for when the workbook is activated.
+    console.log("The workbook was activated.");
```
AlexJerabek, line 370: Should this include the workbook's name?

AlexJerabek (approved):
Approving with suggestions

.gitignore
alison-mk: Yes. This is a MacOS system file that is created whenever one visits a folder in the file explorer UI. I created a workaround for this same file in another repo a few months ago. This is a typical strategy for ensuring that the file isn't tracked in git repos that MacOS users interact with.

docs/excel/excel-add-ins-workbooks.md
alison-mk: The quotation marks are intended to indicate that these words are part of the API name. The [Style Guide](https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=28768) is kind of ambiguous about whether quotation marks are recommended in this type of scenario. I don't want to use bold for emphasis because that usually represents content in the UI (a UI button or worksheet name). I will try switching to italics.

PR #2733, opened 2021-06-05
[Excel] (Table) Add Table.resize article section

alison-mk:
This PR adds an article section for the ExcelApiOnline 1.1 and ExcelApi 1.13 Table.resize API. This API is both a preview API and an online-only API.


docs/excel/excel-add-ins-tables.md
```diff
@@ -160,6 +160,37 @@ Excel.run(function (context) {
 
 ![Table with new calculated column in Excel](../images/excel-tables-add-calculated-column.png)
 
+## Resize a table (preview) (online-only)
+
+> [!NOTE]
+> The `Table.resize` method is currently available in public preview. [!INCLUDE [Information about using preview APIs](../includes/using-excel-preview-apis.md)].
```
AlexJerabek, line 166: Since this is in ExcelApiOnline, it is not in preview and should not have this note. Developers could include this API in production add-ins.

docs/excel/excel-add-ins-tables.md
```diff
@@ -160,6 +160,37 @@ Excel.run(function (context) {
...
+> 
+
+> [!NOTE]
+> The `Table.resize` method is also currently available in ExcelApiOnline 1.1. To learn more, see [Excel JavaScript API online-only requirement set](../reference/requirement-sets/excel-api-online-requirement-set.md).
+
+Your add-in can resize a table without adding data to the table. To resize a table, use the [resize](/javascript/api/excel/excel.table#resize_newRange_) method. The following code sample shows how to resize a table and set the new range of the table to **A1:D20**.
```
AlexJerabek, line 172: Something to consider for a future refactoring (not this PR):
We use this "Your add-in can" phrasing a lot. It's not the most exciting start to a sentence, especially going section by section. Plus, the style guide advises avoiding "can". Maybe we can find a way to improve this.

AlexJerabek, line 172: Suggestions for clarity:
```suggestion
Your add-in can resize a table without adding data to the table or changing cell values. To resize a table, use the [Table.resize](/javascript/api/excel/excel.table#resize_newRange_) method. The following code sample shows how to resize a table and set the new range of the table to **A1:D20**.
```

docs/excel/excel-add-ins-tables.md
```diff
@@ -160,6 +160,37 @@ Excel.run(function (context) {
...
+
+> [!IMPORTANT]
+> The new range of the table must overlap with the original range, and the headers (or the top of the table) must be in the same row.
```
AlexJerabek, line 175: This block breaks the flow between the example explanation to the example. Consider moving it to after the sample.

docs/excel/excel-add-ins-tables.md
```diff
@@ -160,6 +160,37 @@ Excel.run(function (context) {
...
+    // Resize the table.
+    expensesTable.resize("A1:D20");
+
+    return context.sync();
+}).catch(errorHandlerFunction);
+```
+
+**Table after resize** 
+
+![Table with multiple empty rows in Excel](../images/excel-tables-resize.png)
```
AlexJerabek, line 192: Somewhere, we should mention this is the same table as the previous example.

AlexJerabek (changes requested):

docs/excel/excel-add-ins-tables.md
alison-mk: I will reflect on this. I have relied on this phrasing because it's common in our docs, but I agree with your point!

docs/excel/excel-add-ins-tables.md
alison-mk: @ElizabethSamuel-MSFT We discussed offline, and the plan is that I will identify this API as "online-only" here. In the online-only article (to be potentially updated in a subsequent PR), I will consider adding a section that indicates which online-only APIs are candidates for the upcoming GA release.

AlexJerabek (approved):

Issue #2737, opened 2021-06-08
Adding <Icon> element as a child of <Group> raises exception when trying to install the manifest


balazserdeszlogmein:
The documentation says, `<Icon>` is a required element under `<Group>`.
But when adding an ``<Icon> element, the manifest cannot be installed, because it says "only Tooltip and Control are acceptable elements in a Group".

The same is true for `<Title>` under `<Action>` by the way. I suspect these are not available in Outlook add-ins, but the documentation do not mention this information.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 540ad65e-a396-ce57-34fa-00e490028f0d
* Version Independent ID: d7807a56-e4d6-aa64-0dd8-3af677b9849d
* Content: [Group element in the manifest file - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/reference/manifest/group)
* Content Source: [docs/reference/manifest/group.md](https://github.com/OfficeDev/office-js-docs-pr/blob/master/docs/reference/manifest/group.md)
* Product: **non-product-specific**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

AlexJerabek:
Thanks for raising this issue @balazserdeszlogmein.

@Rick-Kirkham, could you please investigate?

Rick-Kirkham:
You are correct. I will update the docs.

PR #2739, opened 2021-06-08
[Outlook] (manifest) specify non-support for some elements in Outlook

Rick-Kirkham:
fixes [2737](https://github.com/OfficeDev/office-js-docs-pr/issues/2737)

### DIFF
--- a/docs/reference/manifest/group.md
+++ b/docs/reference/manifest/group.md
@@ -35,7 +35,10 @@ Required. The label of the group. The **resid** attribute can be no more than 32
 
 ### Icon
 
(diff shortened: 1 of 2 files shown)


docs/reference/manifest/group.md
```diff
@@ -35,7 +35,7 @@ Required. The label of the group. The **resid** attribute can be no more than 32
```
ElizabethSamuel-MSFT, line 35: In the "Child elements" table above, can you note that this doesn't apply to Outlook? Since it says "Icon" is required, readers may assume without going further that it's required for Outlook too.

ElizabethSamuel-MSFT (commented):
You could opt not to convert to notes. However, you should update from "Outlook Add-ins" to "Outlook add-ins".

PR #2745, opened 2021-06-11
[Excel] (Range) Add conceptual docs for getRangeEdge and getExtendedRange

alison-mk:
This PR adds the article sections for the ExcelApi 1.13 getRangeEdge and getExtendedRange APIs.


docs/excel/excel-add-ins-ranges-set-get.md
```diff
@@ -47,6 +47,102 @@ Excel.run(function (context) {
...
+    await context.sync();
+}).catch(errorHandlerFunction);
+```
+
+> [!NOTE]
+> If the selected range is outside a used range, the `Range.getRangeEdge` method will select the cell at the furthest edge of the worksheet in the chosen direction.
+
+#### Before selecting the cell at the edge of the used range
+
+The following screenshot shows a used range and a selected range within the used range. The used range is a table with data at **C5:F12**. Inside this table, the range **D8:E9** is selected. This selection is the *before* state, prior to running the `Range.getRangeEdge` method.
```
AlexJerabek, line 91: I'm on the fence about including these screenshot explanations. I think they do a good job explaining the action, but increase the separation of the images from the code.

docs/excel/excel-add-ins-ranges-set-get.md
```diff
@@ -47,6 +47,102 @@ Excel.run(function (context) {
...
+      activeCell
+    );
+    rangeEdge.select();
+
+    await context.sync();
+}).catch(errorHandlerFunction);
+```
+
+> [!NOTE]
+> If the selected range is outside a used range, the `Range.getRangeEdge` method will select the cell at the furthest edge of the worksheet in the chosen direction.
```
AlexJerabek, line 87: I think there might be a danger of missing cases here, if we try to explain beyond "it's what the keyboard shortcuts do". In this example, Crtl+Arrow either goes to the furthest edge or the start of the next used range. It also has this behavior when the selected range is at the edge of a used range. Plus, there's a complication because the getUsedRange returns a single range, when in reality their can be multiple "used ranges".

So, the two approaches would be to enumerate the full behavior list or mimic the strings in the reference docs that just pass the explanation burden to the UI.

AlexJerabek (changes requested):
Great samples and images. I have some questions around behavior clarification.

docs/excel/excel-add-ins-ranges-set-get.md
alison-mk: I think that, unless you're thoroughly familiar with the UI experience, this concept is so confusing that the screenshots provide a helpful frame of reference. 

Now that the online-only notes are moved, does the content flow feel less broken up?

docs/excel/excel-add-ins-ranges-set-get.md
```diff
@@ -47,6 +47,102 @@ Excel.run(function (context) {
 }).catch(errorHandlerFunction);
 ```
 
+## Select the edge of a used range
+
+An add-in can select the edge of the used range, based on the currently selected range. A used range is any cell or series of contiguous cells that have been edited in the worksheet. In the following screenshot, the used range is the table with values in each cell, **C5:F12**. The empty cells outside this table are outside the used range.
+
+![A table with data from C5:F12 in Excel.](../images/excel-ranges-used-range.png)
```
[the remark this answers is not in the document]
alison-mk: I like that this screenshot provides a baseline for the subsequent two sections, and even while linking to the used range definition, it's still an abstract concept. I think this article section as a whole is more accessible if we define a used range at the start, as it pertains to the subsequent screenshots.  

If we remove the other screenshots then I agree with removing this one, but I think that each of the screenshots provide value by making the concepts more accessible.

docs/excel/excel-add-ins-ranges-set-get.md
AlexJerabek: Agreed that the screenshots are necessary. I was referring to the text blocks (apologies for not being clear):

image

alison-mk: Oh, I see! I included the descriptions explicitly for accessibility, in case a person can't see the screenshot. They're outside the alt text because it's too long for alt text, and it might benefit people who *can* see the screenshot. I'll consult our editor for an additional perspective.

docs/excel/excel-add-ins-ranges-set-get.md
```diff
@@ -47,6 +47,95 @@ Excel.run(function (context) {
 }).catch(errorHandlerFunction);
 ```
 
+## Select the edge of a used range (online-only)
+
+> [!NOTE]
+> The `Range.getRangeEdge` and `Range.getExtendedRange` methods are currently only available in ExcelApiOnline 1.1. To learn more, see [Excel JavaScript API online-only requirement set](../reference/requirement-sets/excel-api-online-requirement-set.md).
+
+The [Range.getRangeEdge](/javascript/api/excel/excel.range#getRangeEdge_direction__activeCell_) and [Range.getExtendedRange](/javascript/api/excel/excel.range#getExtendedRange_directionString__activeCell_) methods let your add-in replicate the behavior of the keyboard selection shortcuts, selecting the edge of the used used range based on the currently selected range. See [Get used range](excel-add-ins-ranges-get.md#get-used-range) to learn more about used ranges.
```
lindalu-MSFT, line 55:
```suggestion
The [Range.getRangeEdge](/javascript/api/excel/excel.range#getRangeEdge_direction__activeCell_) and [Range.getExtendedRange](/javascript/api/excel/excel.range#getExtendedRange_directionString__activeCell_) methods let your add-in replicate the behavior of the keyboard selection shortcuts, selecting the edge of the used range based on the currently selected range. To learn more about used ranges, see [Get used range](excel-add-ins-ranges-get.md#get-used-range).
```
I think the modifier phrase "based on the currently selected range" implies that the UsedRange is in use. "Used used" just doesn't sound right.

docs/excel/excel-add-ins-ranges-set-get.md
```diff
@@ -47,6 +47,95 @@ Excel.run(function (context) {
...
+
+In the following screenshot, the used range is the table with values in each cell, **C5:F12**. The empty cells outside this table are outside the used range.
+
+![A table with data from C5:F12 in Excel.](../images/excel-ranges-used-range.png)
+
+### Select the cell at the edge of the current used range
+
+The following code sample shows how use the `Range.getRangeEdge` method to select the cell at the furthest edge of the current used range, in the direction up. This action matches the result of using the Ctrl+Arrow key keyboard shortcut while a range is selected.
```
lindalu-MSFT, line 63:
```suggestion
The following code sample shows how use the `Range.getRangeEdge` method to select the cell at the furthest edge of the current used range, in the direction up. This action matches the result of using the Ctrl+Up arrow key keyboard shortcut while a range is selected.
```
Seems we should indicate which arrow key to use since the code sample is specifically for the UP direction.

docs/excel/excel-add-ins-ranges-set-get.md
```diff
@@ -47,6 +47,95 @@ Excel.run(function (context) {
...
+    var range = context.workbook.getSelectedRange();
+
+    // Specify the direction with the `KeyboardDirection` enum.
+    var direction = Excel.KeyboardDirection.up;
+
+    // Get the active cell in the workbook.
+    var activeCell = context.workbook.getActiveCell();
+
+    // Get the top-most cell of the current used range.
+    // This method acts like the Ctrl+Arrow key keyboard shortcut while a range is selected.
```
lindalu-MSFT, line 77:
```suggestion
    // This method acts like the Ctrl+Up arrow key keyboard shortcut while a range is selected.
```

docs/excel/excel-add-ins-ranges-set-get.md
lindalu-MSFT: Way to be inclusive! Yes, I agree the text describing the following screenshot is necessary for accessibility.

docs/excel/excel-add-ins-ranges-set-get.md
```diff
@@ -47,6 +47,95 @@ Excel.run(function (context) {
...
+
+#### After selecting the cell at the edge of the used range
+
+The following screenshot shows the same table as the preceding screenshot, with data in the range **C5:F12**. Inside this table, the range **D5** is selected. This selection is *after* state, after running the `Range.getRangeEdge` method to select the cell at the edge of the used range in the up direction.
+
+![A table with data from C5:F12 in Excel. The range D5 is selected.](../images/excel-ranges-used-range-d5.png)
+
+### Select all cells from current range to furthest edge of used range
+
+The following code sample shows how use the `Range.getExtendedRange` method to to select all the cells from the currently selected range to the furthest edge of the used range, in the direction down. This action matches the result of using the Ctrl+Shift+Arrow key keyboard shortcut while a range is selected.
```
lindalu-MSFT, line 102: I mentioned this previously but I'll leave it up to you. I think you should describe which arrow you mean since there are actually four on the keyboard.

From the style guide: Use sentence capitalization to refer to a specific arrow key: the Left arrow key, the Right arrow key, the Up arrow key, or the Down arrow key. It's OK to use arrow key as a general term for any single arrow key. Include the and key in references to a specific arrow key except in key combinations or key sequences.

docs/excel/excel-add-ins-ranges-set-get.md
```diff
@@ -47,6 +47,95 @@ Excel.run(function (context) {
...
+    var range = context.workbook.getSelectedRange();
+
+    // Specify the direction with the `KeyboardDirection` enum.
+    var direction = Excel.KeyboardDirection.down;
+
+    // Get the active cell in the workbook.
+    var activeCell = context.workbook.getActiveCell();
+
+    // Get all the cells from the currently selected range to the bottom-most edge of the used range.
+    // This method acts like the Ctrl+Shift+Arrow key keyboard shortcut while a range is selected.
```
lindalu-MSFT, line 116: Another place to describe which arrow.

docs/excel/excel-add-ins-ranges-set-get.md
```diff
@@ -47,6 +47,95 @@ Excel.run(function (context) {
...
+
+![A table with data from C5:F12 in Excel. The range D8:E9 is selected.](../images/excel-ranges-used-range-d8-e9.png)
+
+#### After selecting all the cells from the current range to the edge of the used range
+
+The following screenshot shows the same table as the preceding screenshot, with data in the range **C5:F12**. Inside this table, the range **D8:E12** is selected. This selection is *after* state, after running the `Range.getExtendedRange` method to select all the cells from the current range to the edge of the used range in the down direction.
+
+![A table with data from C5:F12 in Excel. The range D8:E12 is selected.](../images/excel-ranges-used-range-d8-e12.png)
+
 ## See also
```
lindalu-MSFT, line 139: Note: in this section is a link to the same article. In other words, the link "Set and get the selected range using the Excel JavaScript API" goes to this same article. Recommend removing.

lindalu-MSFT (changes requested):

docs/excel/excel-add-ins-ranges-set-get.md
alison-mk: Oops, that "used used" was a typo. Good catch!

docs/excel/excel-add-ins-ranges-set-get.md
alison-mk: I don't see the issue you describe. The "See also" has a link to a similar article on line 54: "Set and get range values, text, or formulas using the Excel JavaScript API" which links to "excel-add-ins-ranges-set-get-values.md" 

What line is the issue you describe on?

docs/excel/excel-add-ins-ranges-set-get.md
alison-mk: I'll fix this throughout. Thanks!

AlexJerabek (approved):

lindalu-MSFT (approved):

PR #2775, opened 2021-06-30
[Excel] Promote ExcelApi 1.13 requirement set to GA

alison-mk:
This PR: 
- Adds a new ExcelApi 1.13 requirement set article 
- Updates the ExcelApiOnline and Preview articles 
- Removes Preview and Online-only tags from the article sections for APIs released to GA with 1.13

### DIFF
--- a/docs/reference/requirement-sets/excel-api-1-13-requirement-set.md
+++ b/docs/reference/requirement-sets/excel-api-1-13-requirement-set.md
@@ -1,7 +1,7 @@
 ---
 title: Excel JavaScript API requirement set 1.13
 description: 'Details about the ExcelApi 1.13 requirement set.'
-ms.date: 06/30/2021
+ms.date: 07/01/2021
 ms.prod: excel
 localization_priority: Normal
 ---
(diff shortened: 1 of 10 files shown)


docs/reference/requirement-sets/excel-api-1-13-requirement-set.md
```diff
@@ -0,0 +1,57 @@
...
+title: Excel JavaScript API requirement set 1.13
+description: 'Details about the ExcelApi 1.13 requirement set.'
+ms.date: 07/02/2021
+ms.prod: excel
+localization_priority: Normal
+---
+
+# What's new in Excel JavaScript API 1.13
+
+The ExcelApi 1.13 added a method to insert worksheets into a workbook from a base-64 encoded string and an event to detect workbook activation. It also increased support for formulas in ranges by adding APIs to track changes to formulas and locate a formula's direct dependent cells, and it expanded PivotTable support by adding PivotLayout APIs for alt text, style, and empty cell management.
```
ElizabethSamuel-MSFT, line 11: I think the term might be base64 (or Base64). Can you verify?
```suggestion
The ExcelApi 1.13 added a method to insert worksheets into a workbook from a base64-encoded string and an event to detect workbook activation. It also increased support for formulas in ranges by adding APIs to track changes to formulas and locate a formula's direct dependent cells. Additionally, it expanded PivotTable support by adding PivotLayout APIs for alt text, style, and empty cell management.
```

ElizabethSamuel-MSFT (approved):
Looks good. Just a recommendation to split out to separate sentences and confirm if the term is base-64 or base64.

docs/reference/requirement-sets/excel-api-1-13-requirement-set.md
alison-mk: You're right! Good catch. It's a proper noun according to [Term Studio](https://termstudio.azurewebsites.net/viewdata.aspx?tid=777582&langCode=English). I'm not sure how to proceed because our conceptual docs use both base64 and Base64, with many more instances of base64 than Base64. What do you think?

ElizabethSamuel-MSFT: For now, let's update this to Base64. Please open an item on the Content backlog to make this consistent across the docs - whether it be base64 or Base64.

Issue #2777, opened 2021-07-01
invocation code doesn't work.


zhangxj19:
[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: a30fb645-c853-a93a-4137-1e2bf7398a11
* Version Independent ID: 75fb99ac-72e8-7bdc-ff1b-85e7ed930d50
* Content: [Excel custom functions tutorial - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/tutorials/excel-tutorial-create-custom-functions?tabs=excel-windows#create-a-custom-function-that-requests-data-from-the-web)
* Content Source: [docs/tutorials/excel-tutorial-create-custom-functions.md](https://github.com/OfficeDev/office-js-docs-pr/blob/master/docs/tutorials/excel-tutorial-create-custom-functions.md)
* Product: **excel**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

ElizabethSamuel-MSFT:
@zhangxj19 Thanks for letting us know about this. Can you share any additional details about what behavior you're seeing? Also, can you let us know where you're trying this (browser, Windows)?

@alison-mk Can you take a look?

Thanks.

alison-mk:
Hi @zhangxj19, thanks for submitting an issue. I've tested this custom functions tutorial and added a fix for a bug in the *Create a streaming asynchronous custom function* section. It's in PR #2796 now. Once that PR is merged, this issue will automatically close. Let us know if you have any additional questions or concerns. 

Cheers, 
Alison

PR #2796, opened 2021-07-08
[Excel] (Custom Functions) Update custom functions tutorial

alison-mk:
Addressing customer issue #2777.

The customer issue doesn't give any details, but in my own testing of this tutorial I found that the `currentTime()` and `clock()` methods in the *Create a streaming asynchronous custom function* section are automatically populated by yo office for both JS and TS custom function projects, so: 
1) the `currentTime()` and `clock()` methods shouldn't be added to the project again (this causes errors due to duplicate functions) 
2) the project doesn't need to be rebuilt to use **=CONTOSO.CLOCK()**. 

As a result, I adjusted the directions and removed the guidance to re-run `npm run build`.


mscharlock:
The changes look good although this does make me wonder from a larger perspective if we want to write some other functions to explain what is happening since this is simply people running already existing code at this point. Idk what your perspective is - thoughts?

alison-mk:
@mscharlock Could you approve this PR to fix the error in the tutorial, and then I'll track this suggestion as a work item in the custom functions backlog? I agree with your perspective but think it's larger task that will need planning and thought.

mscharlock (approved):

PR #2803, opened 2021-07-09
[Excel] (ribbon) open task panes from contextual tabs

Rick-Kirkham:


docs/design/contextual-tabs.md
```diff
@@ -373,6 +373,91 @@ function myContextChanges() {
...
+                // some control properties omitted
+            }          ]
+        }
+      ]
+    }
+  ]
+}`
+```
+
 ## Localizing the JSON blob
```
lindalu-MSFT, line 461:
```suggestion
## Localize the JSON blob
```
I checked the article, and this is the only heading that uses a gerund verb. Suggest changing to align with other headings in the topic.

lindalu-MSFT (changes requested):
I didn't address the "you cans" in this topic, but there are 10 instances. Suggest you reduce the number of times you use this phrase. For example, in the introductory paragraph:

"You can include custom contextual tabs in your Office Add-in and specify when they are visible or hidden, by creating event handlers that change the visibility."

can be:

"Include custom contextual tabs in your Office Add-in and specify when they are visible or hidden, by creating event handlers that change the visibility."

Rick-Kirkham:
> I didn't address the "you cans" in this topic, but there are 10 instances. Suggest you reduce the number of times you use this phrase. For example, in the introductory paragraph:
> 
> "You can include custom contextual tabs in your Office Add-in and specify when they are visible or hidden, by creating event handlers that change the visibility."
> 
> can be:
> 
> "Include custom contextual tabs in your Office Add-in and specify when they are visible or hidden, by creating event handlers that change the visibility."

I've fixed this in places where it can be. In some places, the "you can" is appropriate because it expresses an option, not something the reader must do.

lindalu-MSFT:
@Rick-Kirkham That's a great way to differentiate. If it's an option, then "you can" is fine. If not, try to write the sentence differently to avoid "you can."

docs/design/contextual-tabs.md
```diff
@@ -373,7 +373,92 @@ function myContextChanges() {
...
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}`
+```
+
+In a task pane, to open a page that is at a different URL from the URL of the shared runtime, specify a `sourceLocation` property in the definition of the action. In the following example, a second task pane is opened from a different button.
```
davidchesnut, line 411:
```suggestion
To open any task pane that is not the default task pane, specify a `sourceLocation` property in the definition of the action. In the following example, a second task pane is opened from a different button.
```

docs/design/contextual-tabs.md
```diff
@@ -373,7 +373,92 @@ function myContextChanges() {
...
+                "label": "Open Tables Taskpane",
+                // some control properties omitted
+            }          ]
+        }
+      ]
+    }
+  ]
+}`
+```
+
```
davidchesnut, line 460: You should also document the `title` and `supportPinning` properties for opening a task pane.
```json
  {
            "id": "showTaskpaneContoso",
            "type": "ShowTaskpane",
            "title": "Contoso task pane",
            "supportPinning": false
          }
```

docs/design/contextual-tabs.md
```diff
@@ -373,7 +373,92 @@ function myContextChanges() {
...
+## Localize the JSON blob
```
davidchesnut, line 461: I think this meant to say JSON. Not a binary large object (BLOB).
```suggestion
## Localize the JSON text
```

docs/design/contextual-tabs.md
```diff
@@ -373,7 +373,92 @@ function myContextChanges() {
...
+    }
+  ]
+}`
+```
+
+In a task pane, to open a page that is at a different URL from the URL of the shared runtime, specify a `sourceLocation` property in the definition of the action. In the following example, a second task pane is opened from a different button.
+
+> [!IMPORTANT]
+>
+> - When a `sourceLocation` is specified for the action, then the task pane does *not* use the shared runtime.
```
davidchesnut, line 415:
```suggestion
> - When a `sourceLocation` is specified for the action, then the task pane does *not* use the shared runtime. It runs in a new JavaScript runtime.
```

davidchesnut (changes requested):
Looking good. A few suggestions. Thanks!

PR #2820, opened 2021-07-20
[All Hosts] (dialog) support for cross-domain messaging

Rick-Kirkham:
Note to reviewers: The two "TBD"s in the requirement set article are intentional. We need to publish before these values are available.

### DIFF
--- a/docs/develop/dialog-api-in-office-add-ins.md
+++ b/docs/develop/dialog-api-in-office-add-ins.md
@@ -125,7 +125,7 @@ Office.context.ui.displayDialogAsync('https://myDomain/myDialog.html', {height: 
 > - The `processMessage` is the function that handles the event. You can give it any name you want.
 > - The `dialog` variable is declared at a wider scope than the callback because it is also referenced in `processMessage`.
 
-The following is a simple example of a handler for the `DialogMessageReceived` event:
+The following is a simple example of a handler for the `DialogMessageReceived` event.
 
 ```js
 function processMessage(arg) {
--- a/docs/develop/dialog-api-in-office-add-ins.md
+++ b/docs/develop/dialog-api-in-office-add-ins.md
@@ -26,7 +26,7 @@ Consider opening a dialog box from a task pane or content add-in or [add-in comm
 
 The following image shows an example of a dialog box.
 
-![Screenshot showing dialog with 3 sign-in options displayed in front of Word](../images/auth-o-dialog-open.png)
+![Screenshot showing dialog with 3 sign-in options displayed in front of Word.](../images/auth-o-dialog-open.png)
 
 Note that the dialog box always opens in the center of the screen. The user can move and resize it. The window is *nonmodal*--a user can continue to interact with both the document in the Office application and with the page in the task pane, if there is one.
 
@@ -36,7 +36,7 @@ The Office JavaScript APIs include a [Dialog](/javascript/api/office/office.dial
 
 To open a dialog box, your code, typically a page in a task pane, calls the [displayDialogAsync](/javascript/api/office/office.ui) method and passes to it the URL of the resource that you want to open. The page on which this method is called is known as the "host page". For example, if you call this method in script on index.html in a task pane, then index.html is the host page of the dialog box that the method opens.
 
-The resource that is opened in the dialog box is usually a page, but it can be a controller method in an MVC application, a route, a web service method, or any other resource. In this article, 'page' or 'website' refers to the resource in the dialog box. The following code is a simple example:
+The resource that is opened in the dialog box is usually a page, but it can be a controller method in an MVC application, a route, a web service method, or any other resource. In this article, 'page' or 'website' refers to the resource in the dialog box. The following code is a simple example.
 
 ```js
 Office.context.ui.displayDialogAsync('https://myAddinDomain/myDialog.html');
@@ -52,7 +52,7 @@ Office.context.ui.displayDialogAsync('https://myAddinDomain/myDialog.html');
 
 After the first page (or other resource) is loaded, a user can use links or other UI to navigate to any website (or other resource) that uses HTTPS. You can also design the first page to immediately redirect to another site.
 
-By default, the dialog box will occupy 80% of the height and width of the device screen, but you can set different percentages by passing a configuration object to the method, as shown in the following example:
+By default, the dialog box will occupy 80% of the height and width of the device screen, but you can set different percentages by passing a configuration object to the method, as shown in the following example.
 
 ```js
 Office.context.ui.displayDialogAsync('https://myDomain/myDialog.html', {height: 30, width: 20});
@@ -67,7 +67,7 @@ Set both values to 100% to get what is effectively a full screen experience. (Th
 
 ### Take advantage of a performance option in Office on the web
 
-The `displayInIframe` property is an additional property in the configuration object that you can pass to `displayDialogAsync`. When this property is set to `true`, and the add-in is running in a document opened in Office on the web, the dialog box will open as a floating iframe rather than an independent window, which makes it open faster. The following is an example:
+The `displayInIframe` property is an additional property in the configuration object that you can pass to `displayDialogAsync`. When this property is set to `true`, and the add-in is running in a document opened in Office on the web, the dialog box will open as a floating iframe rather than an independent window, which makes it open faster. The following is an example.
 
 ```js
 Office.context.ui.displayDialogAsync('https://myDomain/myDialog.html', {height: 30, width: 20, displayInIframe: true});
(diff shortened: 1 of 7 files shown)


docs/develop/dialog-api-in-office-add-ins.md
```diff
@@ -206,6 +211,39 @@ function processMessage(arg) {
...
+> Using `messageParent` to send a cross-domain message requires the [Dialog Origin 1.1 requirement set](../reference/requirement-sets/dialog-origin-requirement-sets.md).
+
+The following is an example of using `messageParent` to send a cross-domain message.
+
+```js
+Office.context.ui.messageParent("Some message", { targetOrigin: "https://target.domain.com" });
+```
+
+> [!NOTE]
+> The `DialogMessageOptions` parameter was released approximately July 19th, 2021. For about 30 days after that date, in Office on the web, the first time that `messageParent` is called without the `DialogMessageOptions` parameter and the parent is a different domain from the dialog, the user will be prompted to approve sending data to the target domain. If the user approves, the user's answer is cached for 24 hours. The user is not prompted again during this period when `messageParent` is called with the same target domain.
```
mscharlock, line 228: What about something like: 
In Office on the web, the user will be prompted to approve sending data to the target domain 30 days after `messageParent` is called without the `DialogMessageOptions` parameter and the parent is a different domain from the dialog?
I'm trying to figure out why the release date is necessary to publish (like, I get that it influences people who already have this code and fit the two conditions) but trying to avoid having to state it directly. 

"The user is not" -> isn't

docs/develop/dialog-api-in-office-add-ins.md
```diff
@@ -206,6 +211,39 @@ function processMessage(arg) {
...
+
+If the message doesn't include sensitive data, you can set the `targetOrigin` to "\*" which allows it to be sent to any domain. The following is an example.
+
+```js
+Office.context.ui.messageParent("Some message", { targetOrigin: "*" });
+```
+
+> [!TIP]
```
mscharlock, line 236: "The simplest way to do this" -> "To do this, create a file..."

mscharlock: I guess if this is a breaking change I would maybe expect this to be something like a warning and not a tip?

mscharlock: Don't know if you need "In the meantime"

mscharlock: Do you need to specify Windows computer? Reason I point this out is that I am unsure how recent announcements of Win365 Cloud PC affect processes like this running virtually but on a Mac.

docs/develop/dialog-api-in-office-add-ins.md
```diff
@@ -293,11 +330,48 @@ function onRegisterMessageComplete(asyncResult) {
...
 
+### Cross-domain messaging to the dialog runtime
+
+The dialog may be navigated away from the add-in's domain or the parent JavaScript runtime (either in a task pane or a UI-less runtime that hosts a function file) may be navigated away from the add-in's domain after the dialog is opened. If either of these things has happened, then a call of `messageChild` will fail unless your code specifies the domain of the dialog runtime. You do this by adding a [DialogMessageOptions](/javascript/api/office/office.dialogmessageoptions) parameter to the call of `messageChild`. This object has a `targetOrigin` property that specifies the domain to which the message should be sent. If the parameter is not used, Office assumes that the target is the same domain that the parent runtime is currently hosting.
```
mscharlock, line 340: Not sure but is it a problem to have sentences worded exactly the same for SEO? Also see previous comments about "The dialog may be navigated away from" sentence.

docs/develop/dialog-api-in-office-add-ins.md
```diff
@@ -293,11 +330,48 @@ function onRegisterMessageComplete(asyncResult) {
...
+dialog.messageChild(messageToDialog, { targetOrigin: "https://target.domain.com" });
+```
+
+If the message doesn't include sensitive data, you can set the `targetOrigin` to "\*" which allows it to be *sent* to any domain. The following is an example.
+
+```js
+dialog.messageChild(messageToDialog, { targetOrigin: "*" });
+```
+
+The JavaScript runtime that is hosting the dialog cannot access the `<AppDomains>` section of the manifest to ensure that the domain *from which the message comes* is trusted, so an alternative has been provided. The object that is passed to the `DialogParentMessageReceived` handler contains the domain that is currently hosted in the dialog as its `origin` property. Your code should use the handler to verify that the message is coming from a trusted domain. The following is an example.
```
mscharlock, line 357: Maybe adjust to something along the lines of 
"Because the JavaScript runtime hosting the dialog can't access the `<AppDomains>` section of the manifest and therefore can't determine if a message's domain is trusted, you must use the `DialogParentMessageReceived` handler to verify trusted messages. The `DialogParentMessageReceieved` handler contains the domain that is currently hosted in the dialog as its `origin` property. The following is an example:"

docs/develop/dialog-api-in-office-add-ins.md
```diff
@@ -293,11 +330,48 @@ function onRegisterMessageComplete(asyncResult) {
...
+        dialog.close();
+        showNotification("Messages from " + arg.origin + " are not accepted.");
+    }
+}
+```
+
+For example, your code could use the [Office.onReady or Office.initialize methods](initialize-add-in.md) to store an array of trusted domains in a global variable. The `arg.origin` property could then be checked against that list in the handler.
+
+> [!TIP]
+> The `DialogMessageOptions` parameter was added to the `messageChild` method as a required parameter in mid-2021. Older add-ins that send a cross-domain message with the method no longer work until they are updated to use the new parameter. In the meantime, *on Office for Windows only*, users and system administrators can enable those add-ins to continue working by specifying the trusted domain(s) with a registry setting. See the **Tip** under [Cross-domain messaging to the host runtime](#cross-domain-messaging-to-the-host-runtime) for details.
```
mscharlock, line 373: Maybe just put the tip in both places? I personally would be annoyed to have to scroll up to re-read a tip (or read it for the first time if I somehow skipped it earlier). Also because this is a breaking change, I think it warrants something other than tip

docs/develop/dialog-best-practices.md
```diff
@@ -36,9 +36,9 @@ For best practices in dialog box design, see [Dialog boxes in Office Add-ins](..
 
 ### Handling pop-up blockers with Office on the web
 
-Attempting to display a dialog box while using Office on the web may cause the browser's pop-up blocker to block the dialog box. Office on the web has a feature that enables your add-in's dialog boxes to be an exception to the browser's pop-up blocker. When your code calls the `displayDialogAsync` method, then Office on the web will open a prompt similar to the following:
+Attempting to display a dialog box while using Office on the web may cause the browser's pop-up blocker to block the dialog box. Office on the web has a feature that enables your add-in's dialog boxes to be an exception to the browser's pop-up blocker. When your code calls the `displayDialogAsync` method, then Office on the web will open a prompt similar to the following.
```
mscharlock, line 39: Maybe combine "Office on the web has a feature" and "When your code calls" for something like 
"However, using the `displayDialogAsync` method will ensure that Office on the web will ensure your dialog boxes are an exception and open a prompt similar to..."

docs/reference/requirement-sets/dialog-origin-requirement-sets.md
```diff
@@ -0,0 +1,41 @@
...
+|:-----|-----|:-----|:-----|:-----|:-----|:-----|:-----|:-----|
+| DialogOrigin 1.1  | Build<br>15.0.5371.1000<br>or later | Build<br>16.0.5200.1000<br>or later | Build<br>TBD<br>or later | TBD | 2.52 or later | 16.52 or later | July, 2021 | Version 2108<br>(Build 10377.1000)<br>or later |
+
+>\* Users of the one-time purchase Office may not have accepted all patches and updates. If so, the DLL that Office uses to report its version in the UI may be greater than the versions listed here even if the updated DLLs needed to support DialogApi have not be installed on the user's computer. To ensure that the needed patch is installed, the user must go to the Office update list ([Office 2013 list](/officeupdates/msp-files-office-2013) or [Office 2016 list](/officeupdates/msp-files-office-2016)), search for **osfclient-x-none**, and install the listed patch.
```
mscharlock, line 19: "have not be installed" -> "have not been installed"

mscharlock, line 19: Is there a specific patch that they need or just generally patches that they might not have that would appear in a list? Maybe adjust to something like:
"To ensure that all necessary patches are installed, go to the Office update list..., and install the missing patches."

mscharlock (commented):
Various comments! Happy to clarify anything. The main thing I am a little confused about is the duplicate instructions - I think I understand why they are necessary but I do wonder if duplicate text has some impact that we should avoid.

docs/develop/dialog-api-in-office-add-ins.md
Rick-Kirkham: The prompt is only going to happen for an approximately 30 day period between July 19th and Aug 19th or thereabouts. We need a way to convey this approximate date range. So, we have to mention the start date.

docs/develop/dialog-api-in-office-add-ins.md
Rick-Kirkham: It's not a breaking change in the usual sense. This is just an optional thing users/admins can do. So, I think is should be a tip.

I think it is understood in the developer world that a virtual machine (e.g. virtualized Windows running in a Mac) counts as a Windows computer.  If it wasn't, then all the world's documentation that mentions an OS-specific thing, would have to be followed by a parenthetical note saying something like "or a virtualized instance running in another host OS".

docs/reference/requirement-sets/dialog-origin-requirement-sets.md
Rick-Kirkham: Actually, that whole footnote should be deleted. It was copied by mistake from another article.

Issue #1137, opened 2019-06-26
There's no way to remove a sideloaded add-in on web or windows


jasonqsong:
In office 365 word web app, after manually uploading the manifest.xml file, the button stick on the ribbon menu and the add-in doesn't shown in the my-addins page

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0ab35128-1e97-c6ad-babc-f696f2e7b0fe
* Version Independent ID: cd1bacb1-14d7-a38d-321c-ebc2a526e03b
* Content: [Sideload Office Add-ins in Office on the web for testing - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/testing/sideload-office-add-ins-for-testing#feedback)
* Content Source: [docs/testing/sideload-office-add-ins-for-testing.md](https://github.com/OfficeDev/office-js-docs-pr/blob/master/docs/testing/sideload-office-add-ins-for-testing.md)
* Product: **non-product-specific**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

jasonqsong:
On Web:

![image](https://user-images.githubusercontent.com/6178947/60190223-e35a2c80-9864-11e9-81f9-0d434e895f6b.png)

![image](https://user-images.githubusercontent.com/6178947/60190294-fec53780-9864-11e9-9e91-0b7c45e68c67.png)

On Windows:

![image](https://user-images.githubusercontent.com/6178947/60190345-113f7100-9865-11e9-8530-84b9f8a087d1.png)

Rick-Kirkham:
It is normal for sideloaded add-ins to not appear in the MY ADD-INS tab. 
This repo is used to track errors in the documentation. For technical puzzles like this, please raise the issue on Stack Overflow and be sure to tag it `office-js`. This will help the whole community benefit from the answers that you get.

andrewkittredge:
This is not a technical puzzle, there is no documented way to uninstall sideloaded add-in.

nilswloewen:
I cannot find any answers to this on stackoverflow either. @andrewkittredge has a legit concern here.

AlexJerabek:
Hi @andrewkittredge and @nilswloewen,

[...]

PR #2845, opened 2021-08-02
[All hosts] (Debugging) Update cache clear article to include web link

alison-mk:
Customer issue #1137 and [internal work item](https://office.visualstudio.com/OC/_workitems/edit/4924870).

### DIFF
--- a/docs/testing/clear-cache.md
+++ b/docs/testing/clear-cache.md
@@ -1,15 +1,18 @@
 ---
 title: Clear the Office cache
 description: 'Learn how to clear the Office cache on your computer.'
-ms.date: 05/22/2020
+ms.date: 08/02/2021
 localization_priority: Priority
 ---
 
 # Clear the Office cache
 
-You can remove an add-in that you've previously sideloaded on Windows, Mac, or iOS by clearing the Office cache on your computer.
+You can remove an add-in that you've previously sideloaded on Windows, Mac, or iOS by clearing the Office cache on your computer. 
 
 Additionally, if you make changes to your add-in's manifest (for example, update file names of icons or text of add-in commands), you should clear the Office cache and then re-sideload the add-in using updated manifest. Doing so will allow Office to render the add-in as it's described by the updated manifest.
+
+> [!NOTE]
+> To remove a sideloaded add-in from Excel, OneNote, PowerPoint, or Word on the web, see [Sideload Office Add-ins in Office on the web for testing: Remove a sideloaded add-in](/sideload-office-add-ins-for-testing.md#remove-a-sideloaded-add-in).
 
 ## Clear the Office cache on Windows
 
--- a/docs/testing/clear-cache.md
+++ b/docs/testing/clear-cache.md
@@ -7,7 +7,7 @@ localization_priority: Priority
 
 # Clear the Office cache
 
-You can remove an add-in that you've previously sideloaded on Windows, Mac, or iOS by clearing the Office cache on your computer. 
+You can remove an add-in that you've previously sideloaded on Windows, Mac, or iOS by clearing the Office cache on your computer.
 
 Additionally, if you make changes to your add-in's manifest (for example, update file names of icons or text of add-in commands), you should clear the Office cache and then re-sideload the add-in using updated manifest. Doing so will allow Office to render the add-in as it's described by the updated manifest.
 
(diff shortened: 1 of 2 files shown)


docs/testing/clear-cache.md
```diff
@@ -11,6 +11,9 @@ You can remove an add-in that you've previously sideloaded on Windows, Mac, or i
 
 Additionally, if you make changes to your add-in's manifest (for example, update file names of icons or text of add-in commands), you should clear the Office cache and then re-sideload the add-in using updated manifest. Doing so will allow Office to render the add-in as it's described by the updated manifest.
 
+> [!NOTE]
+> To remove a sideloaded add-in from Excel, OneNote, PowerPoint, or Word on the web, see [Sideload Office Add-ins in Office on the web for testing: Remove a sideloaded add-in](/sideload-office-add-ins-for-testing.md#remove-a-sideloaded-add-in).
```
AlexJerabek, line 15: The `/` makes the build point to "docs.microsoft.com/sideload-office-add-ins-for-testing" instead of a relative path.
```suggestion
> To remove a sideloaded add-in from Excel, OneNote, PowerPoint, or Word on the web, see [Sideload Office Add-ins in Office on the web for testing: Remove a sideloaded add-in](sideload-office-add-ins-for-testing.md#remove-a-sideloaded-add-in).
```

docs/testing/clear-cache.md
```diff
@@ -11,6 +11,9 @@ You can remove an add-in that you've previously sideloaded on Windows, Mac, or i
 
 Additionally, if you make changes to your add-in's manifest (for example, update file names of icons or text of add-in commands), you should clear the Office cache and then re-sideload the add-in using updated manifest. Doing so will allow Office to render the add-in as it's described by the updated manifest.
 
+> [!NOTE]
```
AlexJerabek, line 14: Would it be better to put this under an H2 instead of a note? That would make it more discoverable in the side navigation. 

image

AlexJerabek (commented):
I'm a little confused by the [sideloading for the web](https://review.docs.microsoft.com/en-us/office/dev/add-ins/testing/sideload-office-add-ins-for-testing?branch=pr-en-us-2845#remove-a-sideloaded-add-in) article. The section linked to here recommends clearing the Office cache on the user's machine, but that shouldn't have an effect, right? Is that an artifact from an old version of the article?

Also, consider having the definitive cache clearing information in this article, instead of in the web sideloading article (and making the web sideloading article link here). I think that makes the reader's page flow a little easier.

docs/testing/clear-cache.md
alison-mk: I waffled about this. At one point, I had a note at the top of the article and also a subsection with an H2. However, I ultimately decided that since this article is titled "Clear the Office cache", and clearing the browser cache is *not* clearing the Office cache, I think including browser cache clear information as an H2 subsection is misleading. As a note at the top of the article, this content has high visibility and directs readers to the appropriate article (if necessary) before digging into Office cache clear steps.

alison-mk:
@AlexJerabek 

> I'm a little confused by the sideloading for the web article. The section linked to here recommends clearing the Office cache on the user's machine, but that shouldn't have an effect, right? Is that an artifact from an old version of the article?

In my experience, clearing the Office cache is an important step when re-sideloading an add-in on the web. @Rick-Kirkham,  can you confirm if the Office cache needs to be cleared after a manifest.xml is changed, before re-uploading that manifest.xml to the web? 

> Also, consider having the definitive cache clearing information in this article, instead of in the web sideloading article (and making the web sideloading article link here). I think that makes the reader's page flow a little easier.

This would involve refactoring multiple articles. I think the change included in this PR addresses the customer feedback. Is there more customer feedback indicating that we should re-think the entire Test and debug TOC section?

docs/testing/clear-cache.md
AlexJerabek: Fair point. I agree with you, though I worry we're favoring technical accuracy over intuitive information discovery. Luckily, we can always watch the feedback (and see the link clicks) to discover if things need to change down the line.

Rick-Kirkham:
> In my experience, clearing the Office cache is an important step when re-sideloading an add-in on the web. @Rick-Kirkham, can you confirm if the Office cache needs to be cleared after a manifest.xml is changed, before re-uploading that manifest.xml to the web?

I'm not aware of any absolute rules about when developers have to clear the Office cache. It was always my understanding that, in theory, it should never be necessary and that it came as a surprise to the engineers a few years ago when it was discovered that sometimes Office didn't recognize changes unless the Office cache was cleared. Clearing the Office cache is a workaround for a bug. (At least that's my understanding.)

alison-mk:
Thanks, Ricky! 

@AlexJerabek, given the information from Ricky, I think leaving the Office cache clear note in the Sideloading for web article is reasonable. Do you think this PR is good to merge, or are there further changes that you'd like to see?

AlexJerabek (approved):

PR #2855, opened 2021-08-05
[Excel] (Custom Functions) Move certificate note

### LINKED ISSUE #2832: Note is in the wrong place in custom functions quick start
The article has a note about installing certificate after the step to run `npm run build`, but the prompt doesn't actually come until after the NEXT step to run `npm run start:desktop`.

image

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4bac6469-78a6-dae9-e6c4-497c953a1576
* Version Independent ID: d860aff2-60ad-c008-2d10-b046f3198f44
* Content: [Custom functions quick start - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/quickstarts/excel-custom-functions-quickstart?tabs=excel-windows)
* Content Source: [docs/quickstarts/excel-custom-functions-quickstart.md](https://github.com/OfficeDev/office-js-docs-pr/blob/master/docs/quickstarts/excel-custom-functions-quickstart.md)
* Product: **excel**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

alison-mk:
Fixes #2832.

### DIFF
--- a/docs/quickstarts/excel-custom-functions-quickstart.md
+++ b/docs/quickstarts/excel-custom-functions-quickstart.md
@@ -53,9 +53,6 @@ To start, you'll use the Yeoman generator to create the custom functions project
     npm run build
     ```
 
-    > [!NOTE]
-    > Office Add-ins should use HTTPS, not HTTP, even when you are developing. If you are prompted to install a certificate after you run `npm run build`, accept the prompt to install the certificate that the Yeoman generator provides.
-
 1. Start the local web server, which runs in Node.js. You can try out the custom function add-in in Excel on the web or Windows. You may be prompted to open the add-in's task pane, although this is optional. You can still run your custom functions without opening your add-in's task pane.
 
 # [Excel on Windows](#tab/excel-windows)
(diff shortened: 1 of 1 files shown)


AlexJerabek (approved):
The change looks good. That said, similar text show up elsewhere in the other quick starts and tutorials. There, the note appears above the command. So, two questions (and tagging @Rick-Kirkham for his opinion too, since he opened the original issue):
- Should these notes come before or after the command?
- Should this note be an include file to reduce the repeated text?

Rick-Kirkham (approved):

Rick-Kirkham:
> The change looks good. That said, similar text show up elsewhere in the other quick starts and tutorials. There, the note appears above the command. So, two questions (and tagging @Rick-Kirkham for his opinion too, since he opened the original issue):
> 
> * Should these notes come before or after the command?
> * Should this note be an include file to reduce the repeated text?

In this case, I think they work best coming just after the command. 
My rule of thumb for include files is that the overhead of maintaining an additional file is worth it only when the include will be used more than 3 times.

AlexJerabek:
@Rick-Kirkham 100% agree with the "3+" rule for include files. I do think in this case that text comes up in every quick start (so about 10 times). Maybe something to put on the backlog?

Rick-Kirkham:
> @Rick-Kirkham 100% agree with the "3+" rule for include files. I do think in this case that text comes up in every quick start (so about 10 times). Maybe something to put on the backlog?

I'll create a work item.

PR #2872, opened 2021-08-12
Quick fix to registry key note

mscharlock:
This was true at time of writing but should be verified by team


millerds (approved):

Rick-Kirkham:
I'm closing this because I already incorporated this change into another PR for the same article.

PR #2877, opened 2021-08-13
[all hosts] (Script Lab) No longer works with IE

ElizabethSamuel-MSFT:
Update alerts in these articles based on info in [Browsers used by Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/concepts/browsers-used-by-office-web-add-ins).


Rick-Kirkham (approved):
I'm approving. Not worth changing now, but I wonder if "[All Hosts]" would have been a more helpful tag than "[admin]" in the PR name.

PR #2894, opened 2021-08-23
[Excel] (Custom Functions) Update web requests article

alison-mk:
This PR adds two cross-links: to the Invocation parameter article section of the parameter options article, and to the invocation object article in the reference docs.


davidchesnut (approved):
Looks good!

PR #2917, opened 2021-09-02
[Excel] (ribbon) fix erroneous OverriddenByRibbonApi support claim

Rick-Kirkham:
`<OverriddenByRibbonApi>` cannot be a child of `<CustomTab>`


docs/design/contextual-tabs.md
```diff
@@ -583,10 +559,10 @@ This simple strategy uses a custom core tab that mirrors a custom contextual tab
...
 
-When a parent tab, group, or menu is marked with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`, then it isn't visible, and all of it's child markup is ignored, when custom contextual tabs aren't supported. So, it doesn't matter if any of those child elements have the `<OverriddenByRibbonApi>` element or what its value is. The implication of this is that if a menu item, control, or group must be visible in all contexts, then not only should it not be marked with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`, but *its ancestor menu, group, and tab must also not be marked this way*.
+When a parent group, or menu is marked with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`, then it isn't visible, and all of it's child markup is ignored, when custom contextual tabs aren't supported. So, it doesn't matter if any of those child elements have the `<OverriddenByRibbonApi>` element or what its value is. The implication of this is that if a menu item or control must be visible in all contexts, then not only should it not be marked with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`, but *its ancestor menu and group must also not be marked this way*.
```
ElizabethSamuel-MSFT, line 562:
```suggestion
When a parent group, or menu is marked with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`, then it isn't visible, and all of its child markup is ignored when custom contextual tabs aren't supported. So, it doesn't matter if any of those child elements have the `<OverriddenByRibbonApi>` element or what its value is. The implication of this is that if a menu item or control must be visible in all contexts, then not only should it not be marked with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`, but *its ancestor menu and group must also not be marked this way*.
```

docs/design/contextual-tabs.md
```diff
@@ -583,10 +559,10 @@ This simple strategy uses a custom core tab that mirrors a custom contextual tab
...
+> Don't mark *all* of the child elements of a group, or menu with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`. This is pointless if the parent element is marked with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>` for reasons given in the preceding paragraph. Moreover, if you leave out the `<OverriddenByRibbonApi>` on the parent (or set it to `false`), then the parent will appear regardless of whether custom contextual tabs are supported, but it will be empty when they are supported. So, if all the child elements shouldn't appear when custom contextual tabs are supported, mark the parent, and only the parent, with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`.
```
ElizabethSamuel-MSFT, line 565:
```suggestion
> Don't mark *all* of the child elements of a group or menu with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`. This is pointless if the parent element is marked with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>` for reasons given in the preceding paragraph. Moreover, if you leave out the `<OverriddenByRibbonApi>` on the parent (or set it to `false`), then the parent will appear regardless of whether custom contextual tabs are supported, but it will be empty when they are supported. So, if all the child elements shouldn't appear when custom contextual tabs are supported, mark the parent, and only the parent, with `<OverriddenByRibbonApi>true</OverriddenByRibbonApi>`.
```

docs/develop/manifest-element-ordering.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: How to find the proper order of manifest elements
 description: 'Learn how to find the correct order in which to place child elements in a parent element.'
-ms.date: 01/29/2021
+ms.date: 02/02/2021
```
ElizabethSamuel-MSFT, line 4:
```suggestion
ms.date: 09/02/2021
```

ElizabethSamuel-MSFT (approved):
Just a few nits. Looks good.

PR #2922, opened 2021-09-03
[All hosts] Update auth namespace to recommended

davidchesnut:
`OfficeRuntime.auth` is deprecated. The correct namespace is now `Office.auth`.


davidchesnut:
Need to postpone PR for now.

PR #2938, opened 2021-09-14
Update tip using F5 to use Ctrl-F5

larsts:
Pressing F5 with add-in in focus will not reload from source, only from cache.
Pressing Ctrl-F5 will reload from source disregarding cache.

I assume hard reload is the most likely needed scenario, as such I propose replacing F5 with Ctrl-F5.
If this assumption is incorrect or not warranted, adding a sentence describing Ctrl-F5 might be a better solution.

### DIFF
--- a/docs/testing/clear-cache.md
+++ b/docs/testing/clear-cache.md
@@ -33,7 +33,7 @@ To remove a sideloaded add-in from Outlook, use the steps outlined in [Sideload 
 Additionally, to clear the Office cache on Windows 10 when the add-in is running in Microsoft Edge, you can use the Microsoft Edge DevTools.
 
 > [!TIP]
-> If you only want the sideloaded add-in to reflect recent changes to its HTML or JavaScript source files, you shouldn't need to clear the cache. Instead, just put focus in the add-in's task pane (by clicking anywhere within the task pane) and then press **F5** to reload the add-in.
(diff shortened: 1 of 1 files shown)


AlexJerabek:
Hi @larsts,

Thanks for submitting this PR. This is a good tip. We are validating that the information is consistent across all the supported Webviews and will respond again soon.

docs/testing/clear-cache.md
```diff
@@ -33,7 +33,7 @@ To remove a sideloaded add-in from Outlook, use the steps outlined in [Sideload
 Additionally, to clear the Office cache on Windows 10 when the add-in is running in Microsoft Edge, you can use the Microsoft Edge DevTools.
 
 > [!TIP]
-> If you only want the sideloaded add-in to reflect recent changes to its HTML or JavaScript source files, you shouldn't need to clear the cache. Instead, just put focus in the add-in's task pane (by clicking anywhere within the task pane) and then press **F5** to reload the add-in.
+> If you only want the sideloaded add-in to reflect recent changes to its HTML or JavaScript source files, you shouldn't need to clear the cache. Instead, just put focus in the add-in's task pane (by clicking anywhere within the task pane) and then press **Ctrl-F5** to reload the add-in.
```
AlexJerabek, line 36: One quick change to match the Microsoft style guide for keyboard command notation.
```suggestion
> If you only want the sideloaded add-in to reflect recent changes to its HTML or JavaScript source files, you shouldn't need to clear the cache. Instead, just put focus in the add-in's task pane (by clicking anywhere within the task pane) and then press **Ctrl+F5** to reload the add-in.
```

AlexJerabek (approved):
Thanks for making this improvement.

PR #2953, opened 2021-09-24
[All Hosts] (manifest) put EquivalentAddin support info in more places

Rick-Kirkham:


docs/reference/manifest/equivalentaddin.md
```diff
@@ -1,14 +1,16 @@
...
 ms.localizationpriority: medium
 ---
 
 # EquivalentAddin element
 
 Specifies backwards compatibility for an equivalent COM add-in or XLL.
 
+[!INCLUDE [Support note for equivalent add-ins feature](../../includes/equivalent-add-in-support-note.md)]
+
 **Add-in type:** Task pane, Custom function
```
ElizabethSamuel-MSFT, line 14:
```suggestion
**Add-in type:** Task pane, Mail, Custom function
```

docs/reference/manifest/equivalentaddins.md
```diff
@@ -1,14 +1,16 @@
...
 ms.localizationpriority: medium
 ---
 
 # EquivalentAddins element
 
 Specifies backwards compatibility with an equivalent COM add-in, XLL, or both.
 
+[!INCLUDE [Support note for equivalent add-ins feature](../../includes/equivalent-add-in-support-note.md)]
+
 **Add-in type:** Task pane, Custom function
```
ElizabethSamuel-MSFT, line 14:
```suggestion
**Add-in type:** Task pane, Mail, Custom function
```

docs/includes/equivalent-add-in-support-note.md
```diff
@@ -0,0 +1,8 @@
+> [!IMPORTANT]
+> The equivalent add-in feature is supported by the following platform and applications, when connected to a Microsoft 365 subscription. COM add-ins cannot be installed on any other platform, so on those platforms the manifest element that is discussed later in this article, `EquivalentAddins`, is ignored.
```
ElizabethSamuel-MSFT, line 2: Since this note is being used in other articles, should link to the original article (Make your Office Add-in compatible...). May also need to consider using include note in other articles but not the original.

Rick-Kirkham: ? I don't understand. This is an include and it is referenced in the other articles where it needs to be.

ElizabethSamuel-MSFT: The current text says, "discussed later in this article" - but most of the includes are being used in articles other than the original. As such, the current text only makes sense in the original article.
```suggestion
> The equivalent add-in feature is supported by the following platform and applications, when connected to a Microsoft 365 subscription. COM add-ins cannot be installed on any other platform, so on those platforms the manifest element `EquivalentAddins` is ignored. For more information, see [Make your Office Add-in compatible with an existing COM add-in](../develop/make-office-add-in-compatible-with-existing-com-add-in.md).
```

Rick-Kirkham: I checked that. The element is "discussed" to some degree in all of the articles where the include is referenced.

ElizabethSamuel-MSFT: Sounds good. I also checked that the "See also" section in the various articles links to the 'original' article.

ElizabethSamuel-MSFT (approved):

PR #2995, opened 2021-10-14
[all hosts] Fix column-header-missing build suggestions

ElizabethSamuel-MSFT:
Can we tweak the -reference script to avoid this issue?


ElizabethSamuel-MSFT:
@alison-mk FYI...

PR #3138, opened 2021-12-15
[All hosts](Web Storage API) add note on Web Storage API availibility

### LINKED ISSUE #3066: On articles about persisting add-in state and settings
The article https://docs.microsoft.com/en-us/office/dev/add-ins/develop/persisting-add-in-state-and-settings seems to be updated recently (09/10/2021) but offers either to save settings within a document (Settings, CustomParts) or to _use techniques provided by browser (localStorage)_.

The issue with localStorage, is that with many modern browsers (namely, Safari and latest Edge builds) even if we use Storage API, this information is not persisted between sessions - I've read similar issues in OfficeDev/office-js (while they were mostly about sharing data between dialog and taskpane) and tested with Safari 15.1, Safari 15.4 and Edge 95: even if `requestStorageAccess` is granted by user (on a click event, as required), for documents in iframe, this permission is reset on browser reload and the worse part is that the localStorage granted to iframe is cleared. Firefox also requires to request a new approval from a user after a browser was closed, but after a new `requestStorageAccess` call the iframe can access all the data from previously used localStorage, so while a user is bothered to click a "Allow Storage"-like button every time the add-in is loaded, it's possible to persist user data and login information in localStorage.

Another relevant article here is https://docs.microsoft.com/en-us/office/dev/add-ins/develop/itp-and-third-party-cookies 
It says:
> Use the Storage Access API so that the third-party can request permission to get access to its first-party cookies. Current versions of Office on Mac and Office on the web both support this API.
> Note
> If you're using cookies for purposes other than authentication, then consider using localStorage for your scenario.

It features nice example (while it does not cover promise rejection), but does not mention that the data won't be preserved.

To my understanding the only solution for Office on the web add-ins would be using `SharedRuntime.storage`.

## Article URL
https://docs.microsoft.com/en-us/office/dev/add-ins/develop/persisting-add-in-state-and-settings
https://docs.microsoft.com/en-us/office/dev/add-ins/develop/itp-and-third-party-cookies

davidchesnut:
This updates the note that Web Storage API (localStorage/sessionStorage) may not be available on some browsers or based on user settings.
Fixes #3066

### DIFF
--- a/docs/develop/persisting-add-in-state-and-settings.md
+++ b/docs/develop/persisting-add-in-state-and-settings.md
@@ -18,7 +18,7 @@ To do that, you can:
 
 - Use techniques provided by the underlying browser control: browser cookies, or HTML5 web storage ([localStorage](https://developer.mozilla.org/docs/Web/API/Window/localStorage) or [sessionStorage](https://developer.mozilla.org/docs/Web/API/Window/sessionStorage)).
     > [!NOTE]
-    > Some browsers, or the user's browser settings may block browser-based storage techniques. You should test for availability as documented in [Using the Web Storage API](https://developer.mozilla.org/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API).
+    > Some browsers or the user's browser settings may block browser-based storage techniques. You should test for availability as documented in [Using the Web Storage API](https://developer.mozilla.org/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API).
 
 This article focuses on how to use the Office JavaScript API to persist add-in state to the current document. If you need to persist state across documents, such as tracking user preferences across any documents they open, you will need to use a different approach. For example, you could use [SSO](sso-in-office-add-ins.md#using-the-sso-token-as-an-identity) to obtain the user identity, and then save the user ID and their settings to an online database.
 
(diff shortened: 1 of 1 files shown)


docs/develop/persisting-add-in-state-and-settings.md
```diff
@@ -18,7 +18,7 @@ To do that, you can:
 
 - Use techniques provided by the underlying browser control: browser cookies, or HTML5 web storage ([localStorage](https://developer.mozilla.org/docs/Web/API/Window/localStorage) or [sessionStorage](https://developer.mozilla.org/docs/Web/API/Window/sessionStorage)).
     > [!NOTE]
-    > The user can block browser-based storage techniques depending on the settings they choose.
+    > Some browsers, or the user's browser settings may block browser-based storage techniques. You should test for availability as documented in [Using the Web Storage API](https://developer.mozilla.org/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API).
```
ElizabethSamuel-MSFT, line 21:
```suggestion
    > Some browsers or the user's browser settings may block browser-based storage techniques. You should test for availability as documented in [Using the Web Storage API](https://developer.mozilla.org/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API).
```

ElizabethSamuel-MSFT (approved):

PR #3145, opened 2021-12-17
[Outlook] (event-based) Include loopback exemption command for https …

ElizabethSamuel-MSFT:
…localhost

Related to issue https://github.com/OfficeDev/office-js/issues/2312.

### DIFF
--- a/docs/outlook/autolaunch.md
+++ b/docs/outlook/autolaunch.md
@@ -2,7 +2,7 @@
 title: Configure your Outlook add-in for event-based activation
 description: Learn how to configure your Outlook add-in for event-based activation.
 ms.topic: article
-ms.date: 12/20/2021
+ms.date: 12/22/2021
 ms.localizationpriority: medium
 ---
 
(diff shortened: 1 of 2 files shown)


ghelyar:
This does not match what I said needed changing to make it work in https://github.com/OfficeDev/office-js/issues/2312

The path to commands.js is wrong and the AppContainer name for `https://localhost:3000/` is wrong, and there's no need to keep `http://localhost` because the rest of the doc only uses `https://localhost:3000/`

ElizabethSamuel-MSFT:
> This does not match what I said needed changing to make it work in [OfficeDev/office-js#2312](https://github.com/OfficeDev/office-js/issues/2312)
> 
> * The path to commands.js is wrong. `src/commands` should be removed from the path (should be `https://localhost:3000/commands.js`)
> * The AppContainer name for `https://localhost:3000/` is wrong, as it needs https, lower case and the guid gets truncated (should be `1_https___localhost_300004aca5ec-d79a-43ea-ab47-e5`)
> * There's no need to keep the old `http://localhost` because the rest of the doc only uses `https://localhost:3000` and what was currently there didn't work anyway.

@ghelyar Thanks for this comment. We'll consider this and update the docs as needed.

Thanks.

ElizabethSamuel-MSFT:
For now, we're updating to truncate the GUID in the command. However, we need to do some further investigation and verification of the exemption and commands items. If further updates are needed, we'll make them in a subsequent PR.

PR #3158, opened 2021-12-28
[excel] (Custom Functions) Add notes about cache clearing to CF quick start and tutorial

AlexJerabek:
Clearing the cache helps avoid a common pitfall in CF development. This PR adds notes about that process early in the user's CF learning journey.


alison-mk (approved):
This is definitely an improvement. I particularly like how it shows a screenshot of a common error message.

PR #3188, opened 2022-01-21
[All hosts] Fix broken sample links

davidchesnut:


AlexJerabek (approved):
All the links on that page work. Thanks for the quick turnaround.

PR #3237, opened 2022-02-11
[Word] (ooxml) Note RichApi as first option

ElizabethSamuel-MSFT:
Related to issue https://github.com/OfficeDev/Office-Add-in-samples/issues/223

### DIFF
--- a/docs/toc.yml
+++ b/docs/toc.yml
@@ -1334,7 +1334,7 @@ items:
     - name: Word object model
       href: word/word-add-ins-core-concepts.md
       displayName: Word
-    - name: Create better add-ins for Word 2013 with Office Open XML
+    - name: Create better add-ins for Word with Office Open XML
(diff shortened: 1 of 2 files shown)


docs/toc.yml
```diff
@@ -1334,7 +1334,7 @@ items:
     - name: Word object model
       href: word/word-add-ins-core-concepts.md
       displayName: Word
-    - name: Create better add-ins for Word with Office Open XML
+    - name: Create better add-ins for Word 2013 with Office Open XML
```
JuaneloJuanelo, line 1337: this is not needed in my view. ALl the content with OOXML is relevant for all versions of office  i would NOT say this is particularly for Office 2013. also this work in Web and Mac.. so saying is 2013 is not only not needed but also misleading.

JuaneloJuanelo (changes requested):

PR #3243, opened 2022-02-14
[all hosts] Update all links to point to GitHub

ElizabethSamuel-MSFT:
Per input from David, update all links to point to GitHub directly instead of the docs samples site.

No longer depends on when PR https://github.com/OfficeDev/Office-Add-in-samples/pull/261 is merged and the Outlook Encrypt Attachments sample indexed.


AlexJerabek (approved):
Looks good. One open question is how to organize the non-application-specific samples. Like, since Shared runtime is not alphabetically before Auth, why does it come earlier in the list? Doesn't need to be solved in this PR, but something to think about.

PR #3265, opened 2022-02-23
[PowerPoint] (Shapes) Modernizing sample code for PPT, OneNote, and missed Excel samples

AlexJerabek:
Part of [this internal work item](https://office.visualstudio.com/OC/_workitems/edit/2747466/) and related to #3254. 

This PR changes more of the samples to use the `async`/`await` pattern to reflect modern JS practices.


Rick-Kirkham (approved):
I'm approving, but most of the "let"s should really be "const"s to conform with best practices. If the variable isn't going to be written to again, it should be a const.

Issue #3296, opened 2022-03-10
Mac cache no com.microsoft folders in ~Library/Containers


barrelhes:
[Enter feedback here]
per the subject - there is no com.microsoft....  folders in ~Library/Containers at all 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4e273e29-10ac-440e-634e-588b5208e484
* Version Independent ID: a5a6dc8a-b504-686e-80ce-45aabd6f832b
* Content: [Clear the Office cache - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/testing/clear-cache)
* Content Source: [docs/testing/clear-cache.md](https://github.com/OfficeDev/office-js-docs-pr/blob/main/docs/testing/clear-cache.md)
* Product: **non-product-specific**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

AlexJerabek:
Hi @barrelhes,

Thank you for raising this issue. @alison-mk, could you please investigate?

alison-mk:
Hi @barrelhes, thanks for submitting this issue. I'm experiencing the same issue as you, so I will investigate internally to see how we should update this article. I'll let you know when I have more information. 

Cheers, 
Alison

alison-mk:
Hi @barrelhes, 

I've located these folders via terminal and via Finder on MacOS. 

To locate via terminal, enter `cd ~/Library/Containers/`, and then enter `ls` to list all the folders in the Containers directory. Use `cd` + folder name to navigate to the correct folder. 

To locate via Finder, make sure to look for Library in the user's root folder, not Library in the system's root folder. Library in the user's root folder is a hidden folder, so you'll need to set Finder to show hidden folders and files to see it. In the Containers directory, the folders are displayed by product name (such as **Microsoft Excel**  instead of **com.microsoft.Excel**).

I've updated the cache clear directions on Mac to clarify this process. This issue will close once that PR is merged. 

Let me know if you have any further questions. 

Thanks, 
Alison

PR #3301, opened 2022-03-12
[All hosts] Update MacOS cache clear instructions

alison-mk:
Answers question asked in #3296.

### DIFF
--- a/docs/includes/mac-cache-folders.md
+++ b/docs/includes/mac-cache-folders.md
@@ -8,12 +8,14 @@ You can clear the cache by using the personality menu of any task pane add-in.
 
     ![Screenshot of clear web cache option on personality menu.](../images/mac-clear-cache-menu.png)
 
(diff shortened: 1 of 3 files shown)


docs/includes/mac-cache-folders.md
```diff
@@ -8,12 +8,14 @@ You can clear the cache by using the personality menu of any task pane add-in.
 
     ![Screenshot of clear web cache option on personality menu.](../images/mac-clear-cache-menu.png)
```
ElizabethSamuel-MSFT, line 9: Is the personality menu available in Outlook Mac? If not, we may want to clarify that in the lead-in statements.

ElizabethSamuel-MSFT (approved):

docs/includes/mac-cache-folders.md
alison-mk: I'm not sure. I will merge this PR and then follow up on this question separately. Created a [work item](https://office.visualstudio.com/OC/_workitems/edit/5898246) to track the task.

PR #3389, opened 2022-04-26
[All Hosts] (function file) add required call of associate method

Rick-Kirkham:


ElizabethSamuel-MSFT:
@Rick-Kirkham In a number of these tutorials, they make an association with `g.methodName = methodName` as well. Are both statements (g...) and the `associate(methodNameInManifest, methodNameInCode)` needed?

Rick-Kirkham:
> @Rick-Kirkham In a number of these tutorials, they make an association with `g.methodName = methodName` as well. Are both statements (`g.method = method` and `associate(methodNameInManifest, methodNameInCode)`) needed?

Fixed.

PR #3402, opened 2022-05-02
[all hosts] (Publishing) Deprecation information

AlexJerabek:
This PR expands our maintainability story by listing our deprecation timeline. This is consistent with the promises made by the Graph team: https://docs.microsoft.com/en-us/graph/versioning-and-support#support-policy-and-deprecation-information.

### DIFF
--- a/docs/publish/maintain-breaking-changes.md
+++ b/docs/publish/maintain-breaking-changes.md
@@ -1,7 +1,7 @@
 ---
 title: Maintain your Office Add-in
 description: Understand our commitments to compatibility and how to keep your add-in up to date.
-ms.date: 04/29/2022
+ms.date: 05/02/2022
 ms.localizationpriority: medium
 ---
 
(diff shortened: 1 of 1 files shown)


docs/publish/maintain-breaking-changes.md
```diff
@@ -30,6 +30,16 @@ If you do encounter an app compatibility issue, App Assure engineers will work w
 
 To learn more about App Assure, watch [Bring your apps to Microsoft Edge with App Assure: tips and tricks](https://techcommunity.microsoft.com/t5/video-hub/bring-your-apps-to-microsoft-edge-with-app-assure-tips-and/ba-p/2167619). To submit your request for app compatibility with App Assure, complete the [Microsoft FastTrack Registration form](https://aka.ms/AppAssureRequest) or send an email to [achelp@microsoft.com](mailto:achelp@microsoft.com).
 
+## Deprecation policy
```
TristanD-MSFT, line 33: What do you think about merging this with the office.js section above? It largely overlaps, so folding them together might read better?

TristanD-MSFT: If we don't merge them, this says we might deprecate over 2 years, that says we generally won't, so I think we need to better align those.

TristanD-MSFT (commented):
Suggest merging the section about office.js breaking changes and the deprecation policy - what do you think?

AlexJerabek:
@TristanD-MSFT, agreed about the merge. The new version combines then a little more cleanly.

samantharamon (approved):
Looks good! Suggesting one fix to the Important alert.

AlexJerabek:
Thanks @samantharamon! I'm waiting for CELA confirmation, then I'll check this in.

Issue #3395, opened 2022-04-29
Invalid example API referenced in page


richstr:
[The following API is referenced in this page, it is no longer valid - https://www.contoso.com/NumberOfPeopleInSpace]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 667d1e98-4a95-318f-0df1-182722370ac8
* Version Independent ID: f8265387-ba9d-a5db-ef44-8409441b2ece
* Content: [Receive and handle data with custom functions - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-web-reqs)
* Content Source: [docs/excel/custom-functions-web-reqs.md](https://github.com/OfficeDev/office-js-docs-pr/blob/main/docs/excel/custom-functions-web-reqs.md)
* Product: **excel**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

AlexJerabek:
Hi @richstr,

The URL used in that sample was intended to be a hypothetical URL. You replace it with a URL that'll be the data source for your scenario.

I'm going to assign this to @alison-mk to either make that more clear or use a real URL.

alison-mk:
Hi @richstr, thanks for submitting this issue. I've updated the article in PR #3403 to address your concern. Once that PR is merged, this issue will close. Let us know if you have any additional questions.

Cheers, 
Alison

richstr:
Thanks!

PR #3403, opened 2022-05-03
[Excel] (Custom functions) Add note to fetch example

alison-mk:
This PR resolves issue #3395 and aligns the article with current style guidelines and practices.

### DIFF
--- a/docs/excel/custom-functions-web-reqs.md
+++ b/docs/excel/custom-functions-web-reqs.md
@@ -7,7 +7,7 @@ ms.localizationpriority: medium
 
 # Receive and handle data with custom functions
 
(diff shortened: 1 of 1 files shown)


AlexJerabek (approved):
Looks good. One general question though, is there a reason we don't want to provide a real, functional URL?

alison-mk:
> One general question though, is there a reason we don't want to provide a real, functional URL?

@AlexJerabek No, not as far as I can tell. I've assigned myself a [work item](https://office.visualstudio.com/OC/_workitems/edit/6027685) to create a functional code sample. I'll merge this PR in the interim.

PR #3426, opened 2022-05-17
[Outlook] (debug) Create debugging article for UI-less add-ins

samantharamon:


timwan10:
This change looks good. Do you think we could put duplicate information or link to here from the UILess/Execute Function section of our documentation too? I'm not sure where the best place would be, I think probably either: 

https://review.docs.microsoft.com/en-us/office/dev/add-ins/outlook/add-in-commands-for-outlook?branch=samramon-onsend-debugging

or 
https://docs.microsoft.com/en-us/office/dev/add-ins/tutorials/outlook-tutorial

? ((probably the first one)

docs/outlook/debug-ui-less.md
```diff
@@ -0,0 +1,39 @@
...
+title: Debug your UI-less Outlook add-in
+description: Learn how to debug your UI-less Outlook add-in.
+ms.topic: article
+ms.date: 05/19/2022
+ms.localizationpriority: medium
+---
+
+# Debug your UI-less Outlook add-in
+
+This article describes how to use the Office Add-in Debugger Extension in Visual Studio Code to debug [UI-less Outlook add-ins](add-in-commands-for-outlook.md#executing-a-javascript-function). UI-less add-in actions are initiated through an add-in command button in the ribbon. For more information about add-in commands, see [Add-in commands for Outlook](add-in-commands-for-outlook.md).
```
lindalu-MSFT, line 11: Do you want to add a link to https://docs.microsoft.com/en-us/office/dev/add-ins/testing/debug-with-vs-extension?

lindalu-MSFT (approved):
Looks good to me!

exextoc:
(timwan) looks good.

docs/outlook/debug-ui-less.md
samantharamon: Thank you for the suggestion. I was considering linking to this article, however, I decided not to add it here as the user may need to perform an additional step before referring to the article. The user may also be running add-ins in the WebView2 runtime, so this article won't apply to them. If the article does apply to the user, they will be directed to it from the 'Configure and run the debugger' section.

ElizabethSamuel-MSFT (approved):

PR #3430, opened 2022-05-19
[Excel] (ExcelApi 1.15) Release ExcelApi 1.15 to GA

alison-mk:
This PR updates a Range article section describing the getDependents method and updates a Table article section to include a description of the new alwaysInsert parameter. 

This PR also removes a number of antiquated ExcelApi 1.2 if statements from the Table article.


AlexJerabek (approved):
Should we add the moniker change to .openpublishing.publish.config.json inthis PR too?

PR #3435, opened 2022-05-20
[admin] Publish

alison-mk:


docs/toc.yml
```diff
@@ -420,6 +418,9 @@ items:
         - name: Concepts (preview)
           href: excel/excel-data-types-concepts.md
           displayName: Excel
+        - name: Entity value cards (preview)
```
ElizabethSamuel-MSFT, line 421: I didn't think we included "(preview)" in the TOC. Did that change?

ElizabethSamuel-MSFT (approved):
Not a blocker so can be addressed in a later update if it turns out to be an issue

docs/toc.yml
alison-mk: I have been including it for the data types articles because the entire feature is still in preview. Open to removing it if that's inconsistent with our process though

alison-mk: Discussed offline, will remove (preview) from TOC in a subsequent commit.

PR #3510, opened 2022-06-28
[All Hosts] (formatting) format element names as bold

Rick-Kirkham:
I'm not changing the dates of the files. This seems like a paradigmatic case when they shouldn't be changed.


docs/tutorials/excel-tutorial.md
```diff
@@ -505,7 +505,7 @@ In this step of the tutorial, you'll add a button to the ribbon that toggles wor
 
 1. Open the manifest file **./manifest.xml**.
 
-1. Locate the `<Control>` element. This element defines the **Show Taskpane** button on the **Home** ribbon you have been using to launch the add-in. We're going to add a second button to the same group on the **Home** ribbon. In between the closing `</Control>` tag and the closing `</Group>` tag, add the following markup.
+1. Locate the **\<Control\>** element. This element defines the **Show Taskpane** button on the **Home** ribbon you have been using to launch the add-in. We're going to add a second button to the same group on the **Home** ribbon. In between the closing **</Control>** tag and the closing **</Group>** tag, add the following markup.
```
alison-mk, line 508: I think these two closing tags need to be escaped, they are missing from the staging article: 
image
Perhaps like this? 
```suggestion
1. Locate the **\<Control\>** element. This element defines the **Show Taskpane** button on the **Home** ribbon you have been using to launch the add-in. We're going to add a second button to the same group on the **Home** ribbon. In between the closing **\</Control\>** tag and the closing **\</Group\>** tag, add the following markup.
```

alison-mk (approved):
Approving with one comment. Once that comment is addressed, this PR is good to go! Thanks for doing this.

PR #3518, opened 2022-07-01
[Excel] (Data types) Add valuesAsJson content to Concepts article

alison-mk:
[Internal work item](https://office.visualstudio.com/OC/_workitems/edit/5869515).


chgrossMSFT (approved):
Looks great to me. Think you did a good job of explaining the basicValue and basicType sections. Tried to find anything to nit but I couldn't!

docs/excel/excel-data-types-concepts.md
```diff
@@ -14,17 +14,30 @@ ms.localizationpriority: high
...
 
-Use the [`Range.valuesAsJson`](/javascript/api/excel/excel.range#excel-excel-range-valuesasjson-member) property to work with data type values. This property is similar to [Range.values](/javascript/api/excel/excel.range#excel-excel-range-values-member), but `Range.values` only returns the four basic types: string, number, boolean, or error values. `Range.valuesAsJson` returns expanded information about the four basic types, and this property can return data types such as formatted number values, entities, and web images.
+The `valuesAsJson` property is integral to creating data types in Excel. This property is an expansion of the `values` property available on multiple Excel JavaScript API objects, like [Range.values](/javascript/api/excel/excel.range#excel-excel-range-values-member). Both the `values` and `valuesAsJson` properties are used to access the value in a cell, but the `values` property only returns one of the four basic types: string, number, boolean, or error values. In contrast, `valuesAsJson` returns expanded information about the four basic types, and this property can return data types such as formatted number values, entities, and web images.
```
AlexJerabek, line 19: Also, I'm not sure `values` returns an error. According to the existing docs, "Cells that contain an error will return the error string."

docs/excel/excel-data-types-concepts.md
```diff
@@ -33,13 +46,17 @@ The `valuesAsJson` property returns a [CellValue](/javascript/api/excel/excel.ce
...
+Each cell value type returned by `valuesAsJson` uses a JSON metadata schema designed for that type. Along with additional properties unique to each data type, these JSON metadata schemas all have the `type`, `basicType`, and `basicValue` properties in common.
+
+The `type` defines the [CellValueType](/javascript/api/excel/excel.cellvaluetype) of the data. The `basicType` is always readonly and is used as a fallback when the data type isn't supported or is formatted incorrectly. The `basicValue` matches the value that would be returned by the `values` property. The `basicValue` is used as a fallback when calculations encounter incompatible scenarios, such as a version of Excel that doesn't meet the minimum build number requirement for the data types feature. The `basicValue` is readonly for `ArrayCellValue`, `EntityCellValue`, `LinkedEntityCellValue`, and `WebImageCellValue` data types.
```
AlexJerabek, line 55: Consider phrasing this with less jargon.
```suggestion
The `type` defines the [CellValueType](/javascript/api/excel/excel.cellvaluetype) of the data. The `basicType` is always readonly and is used as a fallback when the data type isn't supported or is formatted incorrectly. The `basicValue` matches the value that would be returned by the `values` property. The `basicValue` is used as a fallback when calculations encounter incompatible scenarios, such as an older version of Excel that doesn't support the data types feature. The `basicValue` is readonly for `ArrayCellValue`, `EntityCellValue`, `LinkedEntityCellValue`, and `WebImageCellValue` data types.
```

AlexJerabek (approved):
Approving with a couple suggestions.

docs/excel/excel-data-types-concepts.md
alison-mk: Interesting point. It doesn't return an API error (such as `InvalidArgument`), it returns the error value in the cell as a string, such as #NUM!. The data types articles refer to the four basic types this way elsewhere too, with error in the list (see [Data types overview](https://docs.microsoft.com/en-us/office/dev/add-ins/excel/excel-data-types-overview)). 
 
Do you think it would be clearer here if the sentence said "error data type" instead of "error value"?: 

> Both the `values` and `valuesAsJson` properties are used to access the value in a cell, but the `values` property only returns one of the four basic types: string, number, boolean, or error data types.

PR #3521, opened 2022-07-05
[Excel, Word, PPT] (Errors) Abstracting the Excel errors page to app-specific

AlexJerabek:
This PR moves the error code and handling articles to the testing section of the TOC. It also generalizes the Excel error handling article to cover the whole application-specific space. Finally, it adds a note about when you'd get the `NotImplemented` error (which was the genesis of this PR...)

I'm open to thoughts about where these articles should live in the TOC. This is just my proposal.

### DIFF
--- a/docs/testing/application-specific-api-error-handling.md
+++ b/docs/testing/application-specific-api-error-handling.md
@@ -1,21 +1,18 @@
 ---
-title: Error handling with the Excel JavaScript API
-description: Learn about Excel JavaScript API error handling logic to account for runtime errors.
-ms.date: 06/10/2022
+title: Error handling with the application-specific JavaScript APIs
+description: Learn about Excel, Word, PowerPoint, and other application-specific JavaScript API error handling logic to account for runtime errors.
+ms.date: 07/05/2022
 ms.localizationpriority: medium
 ---
 
 
-# Error handling with the Excel JavaScript API
+# Error handling with the application-specific JavaScript APIs
 
-When you build an add-in using the Excel JavaScript API, be sure to include error handling logic to account for runtime errors. Doing so is critical, due to the asynchronous nature of the API.
-
-> [!NOTE]
-> For more information about the `sync()` method and the asynchronous nature of Excel JavaScript API, see [Excel JavaScript object model in Office Add-ins](excel-add-ins-core-concepts.md).
+When you build an add-in using the [application-specific Office JavaScript APIs](../develop/application-specific-api-model.md), be sure to include error handling logic to account for runtime errors. Doing so is critical, due to the asynchronous nature of the APIs.
 
 ## Best practices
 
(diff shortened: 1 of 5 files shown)


Rick-Kirkham (approved):

docs/testing/application-specific-api-error-handling.md
```diff
@@ -64,7 +61,7 @@ The following table is a list of errors that the API may return.
 |`AccessDenied` |You cannot perform the requested operation.| |
 |`ActivityLimitReached`|Activity limit has been reached.| |
 |`ApiNotAvailable`|The requested API is not available.| |
-|`ApiNotFound`|The API you are trying to use could not be found. It may be available in a newer version of Excel. See the [Excel JavaScript API requirement sets](/javascript/api/requirement-sets/excel/excel-api-requirement-sets) article for more information.| |
+|`ApiNotFound`|The API you are trying to use could not be found. It may be available in a newer version of Excel. See the [Excel JavaScript API requirement sets](/javascript/api/requirement-sets) article for more information.| |
```
alison-mk, line 64: Why change the URL but not the link text?

docs/testing/application-specific-api-error-handling.md
```diff
@@ -64,7 +61,7 @@ The following table is a list of errors that the API may return.
 |`AccessDenied` |You cannot perform the requested operation.| |
```
alison-mk, line 61: Are these errors returned in applications other than Excel? About half of the descriptions use Excel-specific language (workbook, worksheet, range, PivotTable, etc) which I think is strange for a page that is supposed to have errors for all application-specific APIs. Maybe we divide the errors into separate tables? Such as: One table for application-neutral, one for Excel only, and one for the other applications if they have any application-specific errors

alison-mk (commented):
I think the new TOC location is good. I have a greater concern about the table containing half/mostly Excel errors. Curious what your thoughts are.

docs/testing/application-specific-api-error-handling.md
AlexJerabek: I didn't fully undo this change...
I wanted to change what page this sends people to, but realized the error text will be in the product.

docs/testing/application-specific-api-error-handling.md
AlexJerabek: I'm torn here. I agree that mixing Excel-specific errors with more general ones is confusing. However, all of these were listed as "Excel" errors, despite also being thrown in Word (and beyond?). To me, this indicates these were Excel-only at some point and have been adopted elsewhere. That's a pattern that could continue. So, rather than having an ever-shifting list or table of which errors can come from which hosts, I think it's easier in the long run to have a main list of app-specific errors.

Put another way, the purpose of this table is as a lookup for when someone gets an error. They see something get thrown, copy and paste the error message into search, land on our page, and scroll through the table. The text surrounding what they're looking doesn't necessarily matter.

Anyway, that's my thoughts.

alison-mk (approved):
Looks good! Thanks for dividing up the error tables.

PR #3550, opened 2022-07-15
[Word] (JS) Replace var with let/const in JS code

samantharamon:
- Updates var in Word JavaScript code.
- Minor formatting fixes.


ElizabethSamuel-MSFT (approved):
Thanks for doing this, including the additional cleanup!

PR #3555, opened 2022-07-19
[Excel] (JS) Replace var with let/const in JS code

samantharamon:
- Updates var in Excel JavaScript code, where applicable.
- Minor formatting fixes.


alison-mk (approved):
Looks good! Thanks for doing this.

PR #3575, opened 2022-08-04
[All Hosts] (browsers) exceptions for WIP-protected documents

Rick-Kirkham:


ElizabethSamuel-MSFT (approved):
Looks good.

Do we have any numbers around WIP-protected documents add-ins are used on? Should we consider adding tabs to the content so e.g., there's are WIP tabs with relevant content in each affected section?

yatibawri1 (approved):

Rick-Kirkham:
> Looks good.
> 
> Do we have any numbers around WIP-protected documents add-ins are used on? Should we consider adding tabs to the content so e.g., there's are WIP tabs with relevant content in each affected section?

Yes. The numbers are very low. That's why Sean wanted us to downplay this instead of making the tables in the article more complex. WIP is already quasi-deprecated. It never caught on is rarely used.

slabereemsft:
WIP Deprecation Announcement - https://docs.microsoft.com/en-us/windows/security/information-protection/windows-information-protection/protect-enterprise-data-using-wip

We're documenting this for completeness and to help with troubleshooting.

PR #3607, opened 2022-08-29
[all hosts] Should be read-only

ElizabethSamuel-MSFT:
Per [style guide](https://styleguides.azurewebsites.net/Styleguide/Read?id=2700&topicid=28837), once it's not the code `readonly`, should be read-only.

Tagging @alison-mk as FYI...


alison-mk (approved):
Thanks!

Issue #3506, opened 2022-06-22
More detailed explanation of Custom Function Limitations


david-ps:
The [limitations list](https://docs.microsoft.com/en-us/office/dev/add-ins/excel/call-excel-apis-from-custom-function#limitations-of-calling-excel-javascript-apis-through-a-custom-function) says custom functions should not: "Add names to a workbook" or "Set properties or execute most methods"

It would be great to know more detail about the implications of these actions and what methods can safely be called and why?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8fe047a0-5f60-f407-2ad0-7f4cc8e52082
* Version Independent ID: f2ebc9bd-25f6-cfc2-dbbe-d0090096f656
* Content: [Call Excel JavaScript APIs from a custom function - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/excel/call-excel-apis-from-custom-function)
* Content Source: [docs/excel/call-excel-apis-from-custom-function.md](https://github.com/OfficeDev/office-js-docs-pr/blob/main/docs/excel/call-excel-apis-from-custom-function.md)
* Product: **excel**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

AlexJerabek:
Hi @david-ps,

Thanks for raising this documentation request. I'll assign this to @alison-mk.

alison-mk:
Hi @david-ps, thank you for submitting this feedback. I will do research internally to determine the answers and then update the article when my research is complete. I'll leave this issue open in the interim.

Cheers,
Alison

alison-mk:
Hi @david-ps, thanks for your patience and for submitting this issue. I've done research internally and have some clarification to offer: 

Custom functions shouldn't change cells outside of the cell running the custom function. Custom functions also should not change the Excel environment (such as by changing Excel's language settings, or by adding worksheets). Custom functions can use Excel JavaScript APIs to *read* information from other cells or the Excel environment, but custom functions shouldn't *write* information to other cells or the Excel environment. 

Using Excel JavaScript APIs to *write* information to other cells or the Excel environment may cause performance issues with the Excel application, such as by causing Excel to crash. If your add-in needs to write information to other cells or the Excel environment, our recommended best practice is to use a task pane or ribbon button.

[...]

PR #3610, opened 2022-08-30
[Excel] (Custom Functions) Further explain custom function limitations

alison-mk:
Expanding description in response to customer issue #3506.

### DIFF
--- a/docs/excel/call-excel-apis-from-custom-function.md
+++ b/docs/excel/call-excel-apis-from-custom-function.md
@@ -52,7 +52,7 @@ A custom functions add-in has the ability to call Excel JavaScript APIs, but you
 - Set properties.
 - Change any of the Excel environment options, such as calculation mode or screen views.
 
-Your custom functions add-in can retrieve information from cells outside the cell running the custom function, but it shouldn't perform write operations to other cells. Instead, make changes to other cells or to the Excel environment from the context of a ribbon button or a task pane. In addition, custom function calculations shouldn't run while an Excel recalculation is taking place, as this will result in unpredictable results.
+Your custom functions add-in can retrieve information from cells outside the cell running the custom function, but it shouldn't perform write operations to other cells. Instead, make changes to other cells or to the Excel environment from the context of a ribbon button or a task pane. In addition, custom function calculations shouldn't run while an Excel recalculation is taking place, as this scenario creates unpredictable results.
 
 ## Next steps
 
--- a/docs/excel/call-excel-apis-from-custom-function.md
+++ b/docs/excel/call-excel-apis-from-custom-function.md
@@ -1,7 +1,7 @@
 ---
 title: Call Excel JavaScript APIs from a custom function
 description: Learn which Excel JavaScript APIs you can call from your custom function.
-ms.date: 07/18/2022
+ms.date: 08/30/2022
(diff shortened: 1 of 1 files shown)


docs/excel/call-excel-apis-from-custom-function.md
```diff
@@ -43,18 +43,16 @@ async function getRangeValue(address) {
 
 ## Limitations of calling Excel JavaScript APIs through a custom function
 
-Don't call Excel JavaScript APIs from a custom function that change the environment of Excel. This means your custom functions should not do any of the following:
+A custom functions add-in has the ability to call Excel JavaScript APIs, but you should be cautious about which Excel JavaScript APIs your add-in calls. Don't call Excel JavaScript APIs from a custom function that change cells outside of the cell running the custom function. Changing other cells or the Excel environment can result in poor performance, time outs, and infinite loops in the Excel application. This means your custom functions shouldn't do any of the following.
```
AlexJerabek, line 46: Suggestions for conciseness.
```suggestion
A custom functions add-in can call Excel JavaScript APIs, but you should be cautious about which APIs it calls. Don't call Excel JavaScript APIs from a custom function that change cells outside of the cell running the custom function. Changing other cells or the Excel environment can result in poor performance, time outs, and infinite loops in the Excel application. This means your custom functions shouldn't do any of the following.
```

docs/excel/call-excel-apis-from-custom-function.md
```diff
@@ -43,18 +43,16 @@ async function getRangeValue(address) {
...
-- Change any of the environment options, such as calculation mode or screen views.
 - Add names to a workbook.
-- Set properties or execute most methods.
+- Set properties.
+- Change any of the Excel environment options, such as calculation mode or screen views.
 
-Changing Excel can result in poor performance, time outs, and infinite loops. Custom function calculations shouldn't run while an Excel recalculation is taking place as it will result in unpredictable results.
-
-Instead, make changes to Excel from the context of a ribbon button, or task pane.
+Your custom functions add-in can retrieve information from cells outside the cell running the custom function, but it shouldn't perform write operations to other cells. Instead, make changes to other cells or to the Excel environment from the context of a ribbon button or a task pane. In addition, custom function calculations shouldn't run while an Excel recalculation is taking place, as this scenario creates unpredictable results.
```
AlexJerabek, line 55: I like using the read/write terminology here to reinforce the underlying actions. Also, conciseness suggestion.
```suggestion
Your custom functions add-in can read information from cells outside the cell running the custom function, but it shouldn't perform write operations to other cells. Instead, make those changes from the context of a ribbon button or a task pane. In addition, custom function calculations shouldn't run while an Excel recalculation is taking place, as this scenario creates unpredictable results.
```

AlexJerabek (approved):
Thanks for making this change. Looks good! Just a couple suggestions to make the writing a little tighter.

docs/excel/call-excel-apis-from-custom-function.md
alison-mk: Thanks for the review and suggestions! I'll take the read/write half of this one, but retain the "changes to other cells or to the Excel environment" phrase to reinforce that the instruction to use a task pane or ribbon button applies to both changes to other cells or to Excel environment.

PR #3618, opened 2022-09-01
[All hosts](Community call) Add latest call blogs

davidchesnut:
Adds the July and August call blogs to the list.


samantharamon (approved):
Looks good!

PR #3643, opened 2022-09-16
[Excel] (Data types) Add new Data types explorer link

alison-mk:
Same addition as https://github.com/OfficeDev/Office-Add-in-samples/pull/395


AlexJerabek (approved):
Looks good, though we should revisit this list to reorganize it now that there are more samples.

Issue #3655, opened 2022-09-24
Please update your instructions


ivose:
https://learn.microsoft.com/en-us/office/dev/add-ins/quickstarts/excel-custom-functions-quickstart?tabs=excel-windows

Here is telling `yo office` And the second step doesn't work:
`Choose a project type: Excel Custom Functions Add-in project` . In my computer it provides no option `Excel Custom Functions Add-in project`. Trying alternatively `Choose a project type="Office Add-in Task Pane project"`, `script type="javascript"`, `name="My custom functions add-in"`, `officeClientApp="Excel"`. Installing.. .
`cd "My office add-in"`, `npm run build`.

And `npm run start:web -- --document https://<myTenantUser>.sharepoint.com/:x:/s/<mysite>/<myExcelUrl>?e=8...g`

After it i upload my own test addins and try excel function `=CONTOSO.GETSTARCOUNT("OfficeDev", "Excel-Custom-Functions")` on b1-cell. It doesn't work.

Commonly please check all of tutorials, because big part of them are obsolete and almost impossible to learn from them.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 4bac6469-78a6-dae9-e6c4-497c953a1576
* Version Independent ID: d860aff2-60ad-c008-2d10-b046f3198f44
* Content: [Custom functions quick start - Office Add-ins](https://learn.microsoft.com/en-us/office/dev/add-ins/quickstarts/excel-custom-functions-quickstart?tabs=excel-windows)
* Content Source: [docs/quickstarts/excel-custom-functions-quickstart.md](https://github.com/OfficeDev/office-js-docs-pr/blob/main/docs/quickstarts/excel-custom-functions-quickstart.md)
* Product: **excel**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

AlexJerabek:
Thanks for reporting this @ivose. I'm sorry to hear you're having issues with the quickstart.

@alison-mk, could you please investigate?

alison-mk:
Hi @ivose, thank you for submitting an issue. We recently updated and renamed the custom functions projects in Yo Office. If you are using Excel on Windows, choose the project type `Excel Custom Functions using a Shared Runtime` and then follow the existing directions. Here is a screenshot: 

[...]

PR #3656, opened 2022-09-27
[Excel] (Custom functions) Update quickstart with new yo office instructions

alison-mk:
Addressing customer issue #3655.


davidchesnut (changes requested):
The image blue text doesn't pass the accessibility image analyzer. Should use a different theme.
Also please update the image for the tutorial as well.

PR #3698, opened 2022-10-13
[Excel] (Custom functions) Remove Script Lab note

alison-mk:
Per eng, the Script Lab step is no longer necessary.


AlexJerabek (approved):
Approving, though it's interesting to me that the SL step isn't necessary anymore but devs will still need to manually add the value in production.

alison-mk:
@AlexJerabek That behavior is true for other custom functions properties too, but I see your point and will pass this feedback along to eng.

PR #3713, opened 2022-10-26
[all hosts] Include missing Word error codes

ElizabethSamuel-MSFT:


alison-mk (approved):
Looks good! Thanks for making these updates.

PR #3719, opened 2022-11-01
[All hosts] (SSO) Fix link

samantharamon:
Fix link that appears in the broken link report.


davidchesnut (approved):
Thanks for catching this!

PR #3749, opened 2022-11-18
[Outlook] (REST) Update REST endpoint decommission note

samantharamon:
The Outlook REST endpoint decommission date has been postponed (previously scheduled for November 30, 2022). As the new decommission date is still pending and the previously announced date is fast approaching, updated the note about the postponement.


AlexJerabek (approved):
Looks good to me!

PR #3751, opened 2022-11-18
[all hosts] (Application-specific APIs) Adding undo behavior note

AlexJerabek:
This PR adds a note to let people know that the undo stack is cleared when an application-specific API is called.

It also does some minor cleanup to the runtimes article, because I considered that article as a place to add this note.

One note for reviewers: I thought about adding information to the custom functions runtime article, but voted against it. Saying non-shared runtimes support undo with CFs sounds too much like an endorsement. However, I'm open to other thoughts.


jeremy-msft (approved):

samantharamon (approved):
Looks good!

Issue #3411, opened 2022-05-06
please add an example returing a dymanic spill array from a streaming function


sdbarrington:
please add an example returing a dymanic spill array from a streaming function

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3aa98c9b-159b-fdc6-03e6-283d483e0d3e
* Version Independent ID: 17f6b1ef-ee34-3f1c-2c9d-828c4a3dd950
* Content: [Return multiple results from your custom function - Office Add-ins](https://docs.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-dynamic-arrays)
* Content Source: [docs/excel/custom-functions-dynamic-arrays.md](https://github.com/OfficeDev/office-js-docs-pr/blob/main/docs/excel/custom-functions-dynamic-arrays.md)
* Product: **excel**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

AlexJerabek:
Hi @sdbarrington,
 
Thank you for the documentation sample request. @alison-mk, can you please investigate and add the request to our backlog as appropriate?

alison-mk:
Hi @sdbarrington, thank you for submitting this request. Here is a GitHub Gist with a code snippet showing a streaming custom function that returns a dynamic spilled array: https://gist.github.com/alison-mk/322b1a3632de8033ead545afbbc8acda. 

I will add this code snippet to our documentation in the near future — that task is in my backlog ([internal tracking link](https://office.visualstudio.com/OC/_workitems/edit/6057195)). I will close this issue now. Let me know if you have any additional questions. 

Cheers, 
Alison

sdbarrington:
thanks!

PR #3759, opened 2022-12-01
[Excel] (Custom functions) Add streaming spill array snippet

alison-mk:
[Internal work item](https://office.visualstudio.com/OC/_workitems/edit/6057195).

Addressing customer issue #3411.

### DIFF
--- a/docs/excel/custom-functions-dynamic-arrays.md
+++ b/docs/excel/custom-functions-dynamic-arrays.md
@@ -1,6 +1,6 @@
 ---
-ms.date: 05/11/2020
-description: Return multiple results from your custom function in an Office Excel add-in.
+ms.date: 11/30/2022
+description: Return multiple results from your custom function in an Excel add-in.
 title: Return multiple results from your custom function
 ms.localizationpriority: medium
 ---
@@ -15,6 +15,8 @@ The following image shows how the `SORT` function spills down into neighboring c
 
 To create a custom function that is a dynamic array formula, it must return a two-dimensional array of values. If the results spill into neighboring cells that already have values, the formula will display a `#SPILL!` error.
 
+## Code samples
+
 The following example shows how to return a dynamic array that spills down.
 
 ```javascript
@@ -28,7 +30,7 @@ function spillDown() {
 }
 ```
 
-The following example shows how to return a dynamic array that spills right.
+The next example shows how to return a dynamic array that spills right.
 
 ```javascript
 /**
@@ -41,7 +43,7 @@ function spillRight() {
 }
 ```
 
-The following example shows how to return a dynamic array that spills both down and right.
+The third example shows how to return a dynamic array that spills both down and right.
 
 ```javascript
 /**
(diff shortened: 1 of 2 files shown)


docs/excel/custom-functions-dynamic-arrays.md
```diff
@@ -58,6 +60,33 @@ function spillRectangle() {
...
+      let thirdResult = 2;
+
+      const timer = setInterval(() => {
+        firstResult += amount;
+        secondResult += amount;
+        thirdResult += amount;
+        invocation.setResult([[firstResult], [secondResult], [thirdResult]]);
+      }, 1000);
+
+      invocation.onCanceled = () => {
```
davidchesnut, line 84: This requires the @cancelable attribute in the code section above. I'm not sure how to test this though to see if it really cancels.

docs/excel/custom-functions-dynamic-arrays.md
```diff
@@ -58,6 +60,33 @@ function spillRectangle() {
 }
 ```
 
+The fourth example shows how to return a dynamic spill array from a streaming function. The results spill down, like the first example, and increment once a second based on the `amount` parameter. To learn more about streaming functions, see [Make a streaming function](custom-functions-web-reqs.md#make-a-streaming-function).
+
+```javascript
+    /** @CustomFunction
```
davidchesnut, line 66:
```suggestion
    /**
 * Increment the cells with a given amount every second. Creates a dynamic spilled array with multiple results
 * @customfunction
 * @param {number} amount The amount to add to the cell value on each increment.
 * @param {CustomFunctions.StreamingInvocation<number[][]>} invocation Parameter to send results to Excel or respond to the user canceling the function. A dynamic array.
 * 
 */
```

Please use this comment section instead for correct case and layout. Also there is no return value, so I moved the description to the top. This makes easier to paste into Script Lab. :-)

docs/excel/custom-functions-web-reqs.md
```diff
@@ -127,6 +126,9 @@ function increment(incrementBy, invocation) {
 }
 ```
 
+> [!NOTE]
```
davidchesnut, line 129: Later in this article is a note that says all cancelable functions must return 1 value. This sample returns void. So I'm not sure if the note needs an update.

davidchesnut (changes requested):
Nice. I added a few suggestions to take a look at. Thanks!

docs/excel/custom-functions-dynamic-arrays.md
alison-mk: My understanding of our documentation is that @cancelable is not relevant for streaming functions: 
image

Screenshot is from this article section: https://learn.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-web-reqs#cancel-a-function

alison-mk: I based this new streaming code sample off the streaming sample in this article: https://learn.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-web-reqs#make-a-streaming-function, which has an onCanceled method and doesn't include @cancelable in the JS doc. Let me know if the existing snippet needs to be updated, too

docs/excel/custom-functions-web-reqs.md
alison-mk: It seems like we each interpret this note differently. Perhaps we can sync offline and then adjust the note accordingly.

alison-mk:
@davidchesnut Thanks for the review! Let's sync offline to discuss further.

docs/excel/custom-functions-dynamic-arrays.md
alison-mk: I clarified this note based on our offline conversation and using information from https://learn.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-json-autogeneration#cancelable

Ready for you to take another look. Thanks!

davidchesnut (approved):
Looks great!

PR #3782, opened 2022-12-16
[all hosts] Refine hub page

ElizabethSamuel-MSFT:


docs/index.yml
```diff
@@ -43,13 +43,13 @@ highlightedContent:
       title: Office Add-ins Beginner's guide
       url: overview/learning-path-beginner.md
     # Card 7
+    - itemType: sample
```
AlexJerabek, line 46: Is there a "Try it out" type?

docs/index.yml
```diff
@@ -81,17 +84,17 @@ conceptualContent:
           text: Develop a full Excel add-in
         - url: tutorials/outlook-tutorial.md
           itemType: tutorial
-          text: Discover the Outlook add-in tutorial
-        - url: overview/patterns-and-practices-overview.md
-          itemType: sample
-          text: Use common Patterns and Practices
+          text: Develop a full Outlook add-in
         - url: /javascript/api/overview
```
AlexJerabek, line 88: Should we add Word here too?

AlexJerabek (approved):
Looks good! I left a couple thoughts for your consideration.

docs/index.yml
ElizabethSamuel-MSFT: unfortunately there isn't. sample seems to be the closest available option

docs/index.yml
ElizabethSamuel-MSFT: added

alison-mk (approved):

PR #3805, opened 2023-01-18
[all hosts] (Platforms) Add note about iOS Conditional Access

AlexJerabek:
This PR adds an explicit note about Conditional Access support for iOS, as this has come up as a question with partners.

### DIFF
--- a/docs/concepts/browsers-used-by-office-web-add-ins.md
+++ b/docs/concepts/browsers-used-by-office-web-add-ins.md
@@ -1,7 +1,7 @@
 ---
 title: Browsers used by Office Add-ins
 description: Specifies how the operating system and Office version determine what browser is used by Office Add-ins.
-ms.date: 12/12/2022
+ms.date: 01/18/2023
 ms.localizationpriority: medium
 ---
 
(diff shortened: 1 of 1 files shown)


Rick-Kirkham (approved):

docs/concepts/browsers-used-by-office-web-add-ins.md
```diff
@@ -42,6 +42,9 @@ For these platforms, the platform alone determines the browser that is used.
 |iOS|any|Safari with WKWebView|
 |Android|any|Chrome|
 
+> [!IMPORTANT]
+> [Conditional Access](/azure/active-directory/conditional-access/overview) is not supported for Office Add-ins on iOS. Those add-ins use embedded Safari, not Edge.
```
jargil, line 46: Could you say WKWebView instead of "embedded Safari"? Or maybe "WKWebView based on Safari".

jargil, line 46: When I click the link in the preview, I get this URL: https://github.com/OfficeDev/office-js-docs-pr/blob/AlexJ-ConditionalAccess/azure/active-directory/conditional-access/overview
I assume this works once the change is published?

AlexJerabek: Yeah, that's a publishing thing. It'll go here: https://learn.microsoft.com/en-us/azure/active-directory/conditional-access/overview

AlexJerabek: Sure. How about "Those add-ins use the Safari-based WKWebView, not an Edge-based browser control."

jargil (approved):

PR #3832, opened 2023-02-03
[all hosts] (Support) Windows Server support notes

AlexJerabek:
This is based on the original question in https://github.com/OfficeDev/office-js-docs-pr/issues/3783. Windows Server 2016+ does support Office Add-ins and WebView2.

This PR also arranges the tables to match the newer-first ordering in [other MSFT documentation](https://learn.microsoft.com/en-us/microsoft-edge/webview2/#supported-platforms). Happy to reverse that change if folks disagree with it. I personally think it helps people understand the current (and most likely) situation quicker.


samantharamon (approved):
Looks good! I agree that the newer-first ordering makes reading the supported client/platform list easier. Would be worth adding an item to the backlog to review our docs to see other articles where we can apply the same format.

PR #3841, opened 2023-02-08
[All hosts] (SSO) add note about port number

davidchesnut:
Add a note to clarify you can use any port (not just 44355).


Rick-Kirkham (changes requested):

davidchesnut:
This issue was taken care of with a different approach in PR #3860

PR #3847, opened 2023-02-13
[all hosts] Remove deprecated 'article' topic

ElizabethSamuel-MSFT:
ms.topic will default to "conceptual"

### DIFF
--- a/docs/outlook/append-on-send.md
+++ b/docs/outlook/append-on-send.md
@@ -1,7 +1,6 @@
 ---
 title: Prepend or append content to a message or appointment body on send
 description: Learn how to prepend or append content to a message or appointment body when the mail item is sent.
-ms.topic: article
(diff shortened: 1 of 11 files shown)


samantharamon (approved):
Thanks for the update. This looks good to me, but just to clarify, will MS Learn automatically add the `ms.topic: conceptual` attribute?

ElizabethSamuel-MSFT:
> Thanks for the update. This looks good to me, but just to clarify, will MS Learn automatically add the `ms.topic: conceptual` attribute?

Thx for asking. In the [docfx for this repo](https://github.com/OfficeDev/office-js-docs-pr/blob/main/docs/docfx.json), we set the default ms.topic as "conceptual". If the file doesn't set/override the ms.topic property, then it'll be "conceptual".

PR #3856, opened 2023-02-18
[All Hosts] (testing) unit testing of collection objects not supported

Rick-Kirkham:


docs/testing/unit-testing.md
```diff
@@ -235,6 +235,9 @@ When you are testing functions that use the application-specific APIs, be sure t
 
 Examples of both types of tests are in the subsections below.
 
+> [!NOTE]
+> The Office-Addin-Mock library doesn't currently support mocking collection type objects. We are working hard on adding this support to the library.
```
alison-mk, line 239: Two minor suggestions: 
1. Per style guide, gerunds are hard to machine translate. 
2. Could you include an example of a collection type object that isn't supported?
```suggestion
> The Office-Addin-Mock library doesn't currently support mocking collection type objects. We are working hard to add this support to the library.
```

alison-mk (approved):
Approved with two small suggestions.

docs/testing/unit-testing.md
Rick-Kirkham:
```suggestion
> The Office-Addin-Mock library doesn't currently support mocking collection type objects, which are all the objects in the application-specific APIs that are named on the pattern *Collection, such as WorksheetCollection. We are working hard to add this support to the library.
```

PR #3862, opened 2023-02-22
Update "Web browser - modern Outlook" blog post link

devantteam:
"Web Browser - modern Outlook" documentation has a broken link to blog post from 2013/09/04 configuring-delegate-access-in-outlook-web-app Current: (relative) /microsoft-365/blog/2013/09/04/configuring-delegate-access-in-outlook-web-app/ Changed to: (relative) /en-us/microsoft-365/blog/2013/09/04/configuring-delegate-access-in-outlook-web-app/

### DIFF
--- a/docs/outlook/delegate-access.md
+++ b/docs/outlook/delegate-access.md
@@ -49,7 +49,7 @@ An Exchange Server feature known as "automapping" is on by default which means t
 
 #### Shared folders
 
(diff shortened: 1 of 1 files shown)


AlexJerabek:
Hi @devantteam,

Good catch. Omitting the `en-us` part of the URL send the reader to the localized version of that content. However, in this case, that blog appears to have never been localized. The link works as is in you're reading in English, but fails otherwise. Tagging @samantharamon for awareness.

AlexJerabek (approved):

Issue #3857, opened 2023-02-20
differentiate from User-Defined Functions


rootsmusic:
How do Custom Functions differ from [User-Defined Functions that are installed with add-ins](https://support.microsoft.com/en-us/office/user-defined-functions-that-are-installed-with-add-ins-reference-729ca23f-e362-4d51-9401-a9060011eb9b?ns=xlwaenduser&version=16&ui=en-us&rs=en-us&ad=us)?  Thanks.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 94ad7803-d4f7-a700-7c7f-204cd8f34e4a
* Version Independent ID: 9c485bae-cab1-bdad-4c8e-baad2b12b22d
* Content: [Create custom functions in Excel - Office Add-ins](https://learn.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-overview)
* Content Source: [docs/excel/custom-functions-overview.md](https://github.com/OfficeDev/office-js-docs-pr/blob/main/docs/excel/custom-functions-overview.md)
* Product: **excel**
* Technology: **add-ins**
* GitHub Login: @o365devx
* Microsoft Alias: **o365devx**

AlexJerabek:
Hi @rootsmusic,

Thank you for the question. UDFs are, effectively, the older, COM-version of custom functions. UDFs don't work on Excel for the web. Custom functions do work cross-platform, as they utilize the Office JS infrastructure.

Let me tag @alison-mk, in case there's a more complete answer or there are places in the docs we could explain this distinction.

rootsmusic:
@AlexJerabek wrote:
> Custom functions do work cross-platform, as they utilize the Office JS infrastructure.

Ok I suggest differentiating the terminology, because "custom functions" also refers to [functions that utilize VBA](https://support.microsoft.com/en-us/office/create-custom-functions-in-excel-2f06c10b-3622-40d6-a1b2-b6748ae8231f) and that can be added as an Excel Add-In.

alison-mk:
Hi @rootsmusic, thank you for the feedback. I've opened a PR to help clarify the **custom functions** terminology in the Office Add-ins documentation. We're also investigating updating the support.microsoft.com articles you linked in your comments, to help reduce confusion throughout our documentation. 

This issue will automatically close when PR #3874 is merged.

Cheers, 
Alison

rootsmusic:
[...]

PR #3874, opened 2023-02-25
[Excel] (Custom functions) Update CF definition based on customer feedback

alison-mk:
Addressing #3857.

### DIFF
--- a/docs/excel/custom-functions-overview.md
+++ b/docs/excel/custom-functions-overview.md
@@ -10,8 +10,7 @@ ms.localizationpriority: high
 
 Custom functions enable developers to add new functions to Excel by defining those functions in JavaScript as part of an add-in. Users within Excel can access custom functions just as they would any native function in Excel, such as `SUM()`.
 
-> [!NOTE]
-> **Custom function** is a general term that is interchangeable with **user-defined function**. Both terms apply to VBA, COM, and Office.js add-ins. The Office Add-ins documentation uses the term **custom function** when referring to custom functions that use Office JavaScript APIs.
+[!include[Excel custom functions definition](../includes/excel-custom-functions-definition.md)]
 
 [!include[Excel custom functions note](../includes/excel-custom-functions-note.md)]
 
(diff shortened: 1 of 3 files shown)


docs/excel/custom-functions-overview.md
```diff
@@ -10,6 +10,9 @@ ms.localizationpriority: high
 
 Custom functions enable developers to add new functions to Excel by defining those functions in JavaScript as part of an add-in. Users within Excel can access custom functions just as they would any native function in Excel, such as `SUM()`.
 
+> [!NOTE]
+> **Custom function** is a general term that is interchangeable with **user-defined function**. Both terms apply to VBA, COM, and Office.js add-ins. The Office Add-ins documentation only uses the term **custom function** when referring to custom functions that use Office.js APIs.
```
AlexJerabek, line 14: I think "only" makes this seem like we have other terms for the broader concept. Also, I believe the term "Office JavaScript APIs" is more common than "Office.js APIs" in our docs.

```suggestion
> **Custom function** is a general term that is interchangeable with **user-defined function**. Both terms apply to VBA, COM, and Office.js add-ins. The Office Add-ins documentation uses the term **custom function** when referring to custom functions that use Office JavaScript APIs.
```

docs/resources/resources-glossary.md
```diff
@@ -67,6 +67,9 @@ A **content delivery network** or **CDN** is a distributed network of servers an
 
 A **custom function** is a user-defined function that is packaged with an Excel add-in. Custom functions enable developers to add new functions, beyond the typical Excel features, by defining those functions in JavaScript as part of an add-in. Users within Excel can access custom functions just as they would any native function in Excel. See [Create custom functions in Excel](../excel/custom-functions-overview.md) to learn more.
 
+> [!NOTE]
+> **Custom function** is a general term that is interchangeable with **user-defined function**. Both terms apply to VBA, COM, and Office.js add-ins. The Office Add-ins documentation only uses the term **custom function** when referring to custom functions that use Office.js APIs.
```
AlexJerabek, line 71:
```suggestion
> **Custom function** is a general term that is interchangeable with **user-defined function**. Both terms apply to VBA, COM, and Office.js add-ins. The Office Add-ins documentation uses the term **custom function** when referring to custom functions that use Office JavaScript APIs.
```

AlexJerabek (approved):
Approving with a couple minor suggestions.

PR #3879, opened 2023-02-27
[Outlook] (attachments) Add limit of Base64-encoded strings

samantharamon:


docs/outlook/add-and-remove-attachments-to-an-item-in-a-compose-form.md
```diff
@@ -34,9 +34,12 @@ You can check for success or error of an asynchronous method call in the callbac
...
+
 ### Attach a file
 
-You can attach a file to a message or appointment in a compose form by using the `addFileAttachmentAsync` method and specifying the URI of the file. You can also use the `addFileAttachmentFromBase64Async` method but specify the base64 string as input. If the file is protected, you can include an appropriate identity or authentication token as a URI query string parameter. Exchange will make a call to the URI to get the attachment, and the web service which protects the file will need to use the token as a means of authentication.
+You can attach a file to a message or appointment in a compose form by using the `addFileAttachmentAsync` method and specifying the URI of the file. You can also use the `addFileAttachmentFromBase64Async` method and specify the Base64-encoded string as input. If the file is protected, you can include an appropriate identity or authentication token as a URI query string parameter. Exchange will make a call to the URI to get the attachment, and the web service which protects the file will need to use the token as a means of authentication.
```
ElizabethSamuel-MSFT, line 42: For your consideration: 1. Keep "but" OR 2. accept the following suggestion

```suggestion
You can attach a file to a message or appointment in a compose form by using the `addFileAttachmentAsync` method and specifying the URI of the file. You can also use the `addFileAttachmentFromBase64Async` method, specifying the Base64-encoded string as input. If the file is protected, you can include an appropriate identity or authentication token as a URI query string parameter. Exchange will make a call to the URI to get the attachment, and the web service which protects the file will need to use the token as a means of authentication.
```

ElizabethSamuel-MSFT (approved):

PR #3880, opened 2023-02-28
Added  "http://" prefix to 'helpUrl' parameter example

CarlosGamez-Nova:
The code example provided for the @helpurl parameter does not use the "http://" prefix. If the developer implements it like this it won't work. The automatically created functions.json file does not include the http prefix and Excel fails to recognise it when clicking on it on the desktop app.

The documentation for manually creating the functions.json file correctly indicates the use of the "http://" prefix. See https://learn.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-json#json-metadata-example

### DIFF
--- a/docs/excel/custom-functions-json-autogeneration.md
+++ b/docs/excel/custom-functions-json-autogeneration.md
@@ -135,7 +135,7 @@ Syntax: @helpurl *url*
 
 The provided *url* is displayed in Excel.
 
-In the following example, the `helpurl` is `www.contoso.com/weatherhelp`.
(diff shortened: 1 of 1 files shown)


ElizabethSamuel-MSFT:
@CarlosGamez-Nova Thanks for creating this PR.

Tagging @alison-mk to review.

Thanks.

alison-mk:
Hi @CarlosGamez-Nova, thank you for submitting this PR! I will merge your PR. We also need to change the @helpurl text in line 144 (inside the code snippet). I will do that in a subsequent PR after merging yours. 

Cheers, 
Alison

PR #3885, opened 2023-03-02
[All hosts] (SSO) Update steps

samantharamon:
- Updates some Azure Portal navigation steps.
- Applies minor formatting fixes.


davidchesnut (approved):
I must not have had my coffee when I last updated this. Thanks for catching all these edit mistakes!

PR #3889, opened 2023-03-03
[all hosts] (CDN) Information on the China-specific office.js CDN

AlexJerabek:
To stay compliant and performant within the [21Vianet cloud](https://learn.microsoft.com/en-us/microsoft-365/admin/services-in-china/services-in-china?view=o365-21vianet), LOB add-ins specific to that region should use the CDN designed for that region. This PR adds that information to two strategic spots in the documentation.


docs/develop/referencing-the-javascript-api-for-office-library-from-its-cdn.md
```diff
@@ -44,7 +44,12 @@ New JavaScript APIs are first introduced in "preview" and later become part of a
 
 [!INCLUDE [Information about using preview APIs](../includes/using-preview-apis-host.md)]
 
+## CDN references for other M365 environments
```
ElizabethSamuel-MSFT, line 47:
```suggestion
## CDN references for other Microsoft 365 environments
```

ElizabethSamuel-MSFT (approved):
We mention the default CDN in a number of places. Do we need to point out or link to this alternate in those places?

mattgeim (approved):

AlexJerabek:
> We mention the default CDN in a number of places. Do we need to point out or link to this alternate in those places?

I think we're good strategically referencing and linking to this CDN. We don't want folks using this CDN otherwise of China.

Issue #3904, opened 2023-03-12
appendOnSendAsync() and prependOnSendAsync() methods don't work


elegault:
My assumption after reading the article was that a specified function would run _automatically_ when the add-in is sent. However, the two sample functions are connected to Add-in commands associated with two Ribbon buttons, which must be clicked by the user to call the relevant `Office.context.mailbox.item.body.appendOnSendAsync` or `Office.context.mailbox.item.body.prependOnSendAsync` functions. Clicking these buttons show informational messages, but the messages don't complete or close and nothing is appended or prepended in the email body after the email is sent (in Outlook for Windows Version 2303 (Build 16227.20004 Click-to-Run).

![image](https://user-images.githubusercontent.com/4130365/224578031-f66f7e39-0d9a-430f-8a96-b8fab630f25b.png)

Is event-based activation with the OnMessageSend event the only way we can automatically run those two functions without user interaction? I tried implementing the WebView and JS runtimes as per event-based activation requirements to see if that would get the functions to work, but it didn't work. If those runtimes aren't needed, then the article should remove the references to WebViewRuntime.Url and JSRuntime.Url in the manifest snippet to alleviate any confusion.

Note also that there is a reference to a commands.js file, but if the user follows the instructions as per [the quick start sample](https://learn.microsoft.com/en-us/office/dev/add-ins/quickstarts/outlook-quickstart?tabs=yeomangenerator) that is referenced, that sample does not have a commands.js file!

I have a full sample project hosted in GitHub Pages in my repo which is based on this article: https://github.com/elegault/PnP-OfficeAddins/tree/outlook-prepend-or-append-on-send/Samples/outlook-prepend-or-append-on-send.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 28320183-499e-5a19-2925-31f73227766f
* Version Independent ID: e998c64b-dc57-fb8d-9959-c4095e32411a
* Content: [Prepend or append content to a message or appointment body on send - Office Add-ins](https://learn.microsoft.com/en-us/office/dev/add-ins/outlook/append-on-send?tabs=xmlmanifest)
* Content Source: [docs/outlook/append-on-send.md](https://github.com/OfficeDev/office-js-docs-pr/blob/main/docs/outlook/append-on-send.md)
* Product: **outlook**
* Technology: **add-ins**
* GitHub Login: @o365devx
[...]

PR #3913, opened 2023-03-15
[Outlook] (prepend-on-send) Clarify feature description

samantharamon:
Based on feedback in #3904:
- Removes `WebViewRuntime.Url` and `JSRuntime.Url` resources from the manifest, as these aren't required for the walkthrough to work.
- Clarifies how the `prependOnSendAsync` and `appendOnSendAsync` methods are activated.

### DIFF
--- a/docs/outlook/append-on-send.md
+++ b/docs/outlook/append-on-send.md
@@ -207,6 +207,9 @@ The following shows how to configure your JSON manifest to enable the append-on-
 ---
 
 > [!TIP]
+> The prepend-on-send and append-on-send features must be activated by the user through a task pane or function command button. If you want content to be prepended or appended on send without additional action from the user, you can implement these features in an [event-based activation add-in](autolaunch.md).
+
+> [!TIP]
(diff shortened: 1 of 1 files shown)


docs/outlook/append-on-send.md
```diff
@@ -350,6 +350,9 @@ In this section, you'll implement the JavaScript code to append a sample company
 
     ![A sample of a sent message with the Contoso header prepended and the disclaimer appended to its body.](../images/outlook-prepend-append-on-send.png)
 
+    > [!TIP]
+    > Because content is only prepended or appended once the message is sent, the sender will only be able to view the added content from their Inbox or Sent Items folder. If you require the sender to view the added content before the message is sent, see [Insert data in the body when composing an appointment or message in Outlook](insert-data-in-the-body.md).
```
ElizabethSamuel-MSFT, line 354:
```suggestion
    > Because content is only prepended or appended once the message is sent, the sender will only be able to view the added content from their **Inbox** or **Sent Items** folder. If you require the sender to view the added content before the message is sent, see [Insert data in the body when composing an appointment or message in Outlook](insert-data-in-the-body.md).
```

PR #3927, opened 2023-03-20
[Excel] (Custom functions) Refresh docs to include shared runtime yo office project

alison-mk:
[Work item.](https://office.visualstudio.com/OC/_workitems/edit/6154073)

This PR updates our docs so that customers are consistently guided to use the **Excel Custom Functions using a Shared Runtime -- JS** project type. Custom functions without a shared runtime are treated as a legacy project type. 

It also removes a series of steps from the **Share custom function data with the task pane tutorial** because those steps are: 
1. Irrelevant for customers that are following our suggested path of using the shared runtime yo office project, and 
2. Described elsewhere in our docs, in the [Configure your Office Add-in to use a shared runtime](https://learn.microsoft.com/en-us/office/dev/add-ins/develop/configure-your-add-in-to-use-a-shared-runtime) article


docs/tutorials/share-data-and-events-between-custom-functions-and-the-task-pane-tutorial.md
```diff
@@ -1,158 +1,18 @@
...
-   ```command&nbsp;line
-   npm run start
-   ```
-
-## Share state between custom function and task pane code
-
-Now that custom functions run in the same context as your task pane code, they can share state directly without using the **Storage** object. The following instructions show how to share a global variable between custom function and task pane code.
+The following instructions show how to share a global variable between custom function and task pane code. This tutorial assumes that you've completed the [Excel custom functions tutorial](excel-tutorial-create-custom-functions.md), with a **Excel Custom Functions using a Shared Runtime** project using the script type **JavaScript**. Use the add-in you created in that tutorial to complete the following instructions.
 
 ### Create custom functions to get or store shared state
```
davidchesnut, line 17: The step at line 199 can be removed (adding the script to include functions.js in the html file). Since yo office uses webpack this already happens automatically.

docs/tutorials/excel-tutorial-create-custom-functions.md
```diff
@@ -221,4 +221,4 @@ You can prevent this conflict by [clearing the Office cache](../testing/clear-ca
 
 ## Next steps
```
davidchesnut, line 222: Maybe another bug/PR to track, but adding the clock function in the previous steps causes an error. The yo office project already has this code. so we need a different streaming example (or perhaps take this one out?)

davidchesnut (changes requested):
This is great! A few comments added. There were a couple of issues that perhaps should be separate PRs or bugs (as I don't know if they are in the scope of this PR) so please check. Thanks!

docs/develop/yeoman-generator-overview.md
```diff
@@ -39,14 +39,16 @@ yo office
...
 - **Office Add-in Task Pane project using React framework**
+- **Office Add-in Task Pane project using Angular framework**
+- **Excel Custom Functions using a Shared Runtime**
+- **Excel Custom Functions using a JavaScript-only Runtime**
 - **Office Add-in Task Pane project supporting single sign-on**
+- **Office Add-in with Teams Manifest (preview)**
 - **Office Add-in project containing the manifest only**
-- **Excel Custom Functions Add-in project**
 
 ![Screenshot showing the prompt for project type, and the possible answers, in the Yeoman generator.](../images/yo-office-project-type-prompt.png)
```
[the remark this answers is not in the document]
alison-mk: Thanks for reviewing! 

The manifest only project is still available, but it's below the fold — at the bottom of the screenshot you see the note "Move up and down to see more choices". Moving up or down reveals the manifest only project option. 

I'll update the note to remove the SSO option, but leave the manifest only information in the note and on line 83.

docs/tutorials/share-data-and-events-between-custom-functions-and-the-task-pane-tutorial.md
alison-mk: Fixed! Thanks

docs/tutorials/excel-tutorial-create-custom-functions.md
alison-mk: Thanks! Created a [work item](https://office.visualstudio.com/OC/_workitems/edit/7718962) to track this task separately.

alison-mk (commented):
Thanks for reviewing @davidchesnut! I incorporated your feedback, ready for a second look.

davidchesnut (approved):
Looks great!

PR #3965, opened 2023-04-18
[all hosts] (SSO) Fix example API link text

AlexJerabek:
The "for example" text in this include file seems like it should use an example, rather than the placeholder text.


davidchesnut (approved):
Good catch! Thanks!

PR #4153, opened 2023-08-30
Update js file source to be correct one

tomaskukk:
The docs request one to create `spamreporting.js` file that will be used as the entry point for the event handling, but the `xml` file still has the "old" entry point.


samantharamon:
Hi @tomaskukk,

Thanks for testing the integrated spam-reporting feature and catching this doc bug! Merging the change.

PR #4216, opened 2023-10-13
[all hosts] Fix build suggestions

ElizabethSamuel-MSFT:
Also, make links to ask questions and request features more consistent.

### DIFF
--- a/Issue_Template.md
+++ b/Issue_Template.md
@@ -5,7 +5,7 @@ To report an issue with the Office-Add-ins documentation, please provide the art
 
 If your issue is not related to the Office Add-ins documentation, please post it to one of the following channels instead.
 
-- To ask a question about using the Office.js API, post your question to Microsoft Q&A (https://learn.microsoft.com/answers/tags/321/office-development).
+- To ask a question about using the Office.js API, post your question to Stack Overflow and tag it with the "office-js" tag (https://stackoverflow.com/questions/tagged/office-js).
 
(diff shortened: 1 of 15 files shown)


Issue_Template.md
```diff
@@ -5,11 +5,11 @@ To report an issue with the Office-Add-ins documentation, please provide the art
 
 If your issue is not related to the Office Add-ins documentation, please post it to one of the following channels instead.
 
-- To ask a question about using the Office.js API, post your question to Stack Overflow and tag it with the "office-js" tag (http://stackoverflow.com/questions/tagged/office-js).
+- To ask a question about using the Office.js API, post your question to Microsoft Q&A (https://learn.microsoft.com/answers/tags/321/office-development).
```
AlexJerabek, line 8: I thought we were favoring sending folks to Stack since it's more likely to get them an answer.

ElizabethSamuel-MSFT: I'll make this change in the issue template. Should I also update in the application-specific index.yml files?

AlexJerabek: I'm not sure what the right call is. I'm very biased to Stack since it's an industry standard. I remember Linda pushing for Q&A to be the standard. Our [official resources page](https://learn.microsoft.com/en-us/office/dev/add-ins/resources/resources-links-help?view=excel-js-preview) lists both, but leads with Stack.

To your direct question, I think the index pages and the issue template should point to the same place.

AlexJerabek (approved):

PR #4231, opened 2023-10-20
[WXP] (Application-specific model) Add a section on loading properties from a collection

AlexJerabek:
I couldn't find this information in the docs. Since this came up in an internal thread, I'm adding a small section about loading from a collection. If I'm mistaken and this information exists elsewhere (or I've oversimplified something), please let me know.

### DIFF
--- a/docs/develop/application-specific-api-model.md
+++ b/docs/develop/application-specific-api-model.md
@@ -148,7 +148,7 @@ Please be aware that some of the properties under an object may have the same na
 
 #### Load from a collection
 
-When working with a collection, use `load` on the collection to load properties for every object in the collection. Use `load` exactly as you would for an individual object in that collection. Don't include the `items` property of the collection in the `load` arguments.
+When working with a collection, use `load` on the collection to load properties for every object in the collection. Use `load` exactly as you would for an individual object in that collection.
 
 The following sample code shows the `name` property being loaded and logged for every chart in the "Sample" worksheet.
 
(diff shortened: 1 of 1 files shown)


docs/develop/application-specific-api-model.md
```diff
@@ -146,6 +146,27 @@ You can also set the scalar properties of a navigation property by traversing th
 
 Please be aware that some of the properties under an object may have the same name as another object. For example, `format` is a property under the `Excel.Range` object, but `format` itself is an object as well. So, if you make a call such as `range.load("format")`, this is equivalent to `range.format.load()` (an undesirable empty `load()` statement). To avoid this, your code should only load the "leaf nodes" in an object tree.
 
+#### Load from a collection
+
+When working with a collection, use `load` on the collection to load properties for every object in the collection. Use `load` exactly as you would for an individual object in that collection. Don't include the `items` property of the collection in the `load` arguments.
```
Rick-Kirkham, line 151: Script Lab won't load for me today, but I *think* the following change is true. Can you test on your computer?

```suggestion
When working with a collection, use `load` on the collection to load properties for every object in the collection. Use `load` exactly as you would for an individual object in that collection. You normally don't include the `items` property of the collection in the `load` arguments, because all the items are loaded if you load any item properties. However, if you will be looping over the items in the collection, but don't need to load any particular property of the items, you need to `load` the `items` property.
```

Rick-Kirkham (changes requested):

docs/develop/application-specific-api-model.md
AlexJerabek: Good catch. I verified this in Script Lab. I'm using your text, plus a small sample.

Rick-Kirkham (approved):

PR #4232, opened 2023-10-23
[Outlook] (Smart Alerts) Change implemented send mode option in walkthrough

samantharamon:
Based on feedback, changes implemented send mode option in walkthrough to **soft block**.


AlexJerabek (approved):
Changes seem good. Just want to check that the reference to (non-soft) "block" in the "Override the send mode option at runtime" section is intentional.

samantharamon:
> Changes seem good. Just want to check that the reference to (non-soft) "block" in the "Override the send mode option at runtime" section is intentional.

Yes, that's intentional to provide a sample scenario.

PR #4233, opened 2023-10-23
[Excel] (Custom functions) Updating equivalent XLL article

alison-mk:
Editing pass of article to remove future tense and clarify some confusing sentences. Also updating TOC title to make it more clear what the article content contains.


davidchesnut (approved):
Nice! Thanks for refreshing this!

PR #4235, opened 2023-10-26
[all hosts] Rework hub page to be more scenario/intent focused 

AlexJerabek:
Our hub page has higher-than-expected bounce rates and exit rates. This PR removes some of the links to pages with lower clicks. and promotes some scenarios, with the intention to substitute newer scenarios as they come online.

### DIFF
--- a/docs/index.yml
+++ b/docs/index.yml
@@ -32,18 +32,19 @@ highlightedContent:
       url: overview/office-add-in-code-samples.md
 
 productDirectory:
+  title: Get started with Office Add-ins # < 60 chars (optional)
   items:
-  - title: Get started in 5 minutes
+  - title: 5-minute quick starts
     imageSrc: images/index-landing-page/i_quick-start.svg
     links:
       - url: quickstarts/excel-quickstart-jquery.md
-        text: Excel add-in
+        text: Build an Excel task pane add-in
       - url: quickstarts/outlook-quickstart.md
-        text: Outlook add-in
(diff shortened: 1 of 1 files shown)


docs/index.yml
```diff
@@ -23,34 +23,49 @@ highlightedContent:
...
-      title: Test and experiment with APIs using Script Lab
-      url: overview/explore-with-script-lab.md
-    # Card 8
-    - itemType: reference
-      title: Office JavaScript API reference documentation
-      url: /javascript/api/overview
+      title: Samples for Office Add-ins
+      url: overview/office-add-in-code-samples.md
 
+productDirectory:
```
samantharamon, line 34: I understand that the `title` attribute is [optional](https://review.learn.microsoft.com/help/platform/navigation-how-create-hub-page?branch=main#product-directory-section-multi-link), but should we consider adding one to the `productDirectory` section to match the `conceptualContent` section? Apart from the addition of the icon to the `productDirectory` cards, these sections share the same format (versus that of the top-level section). Not a deal-breaker, but thought it was worth asking.

samantharamon (approved):
Changes look good! Added a minor question on formatting.

docs/index.yml
AlexJerabek: Good point. I agree that a title there would help draw attention. I'll try to think of something fitting, but I'm open to suggestions.

davidchesnut (approved):
This looks much better focused!

PR #4236, opened 2023-10-26
[All hosts] Fix broken link

samantharamon:


AlexJerabek (approved):
Frustrating that we have to include a loc parameter. Thanks for fixing this though.

PR #4267, opened 2023-11-10
[all hosts] (TOC) Remove Samples node from TOC

AlexJerabek:
This PR adjusts the Samples node in the TOC. Previously, a few samples existed there, but it didn't represent the full breadth of our sample library and may have given false impressions. The existing sample pages have been moved elsewhere in the TOC.

This PR also makes a tighter connection between the Script Lab article and the sample to convert a Script Lab snippet into an add-in.

### DIFF
--- a/docs/toc.yml
+++ b/docs/toc.yml
@@ -32,6 +32,9 @@ items:
   - name: Explore Office JS APIs using Script Lab
     href: overview/explore-with-script-lab.md
     displayName: Excel, Word, PowerPoint
+  - name: Samples
+    href: overview/office-add-in-code-samples.md
+    displayName: PnP
   - name: Set up your development environment 
     href: overview/set-up-your-dev-environment.md
     displayName: Excel, Word, OneNote, Project, PowerPoint, Outlook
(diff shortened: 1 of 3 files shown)


docs/toc.yml
```diff
@@ -32,6 +32,9 @@ items:
   - name: Explore Office JS APIs using Script Lab
     href: overview/explore-with-script-lab.md
     displayName: Excel, Word, PowerPoint
+  - name: Samples
+    href: overview/office-add-in-code-samples.md
+    displayName: PnP
```
samantharamon, line 37: Should we add additional terms here to ensure that it appears as a result during TOC filtering, e.g., "code" or "code samples"?

docs/toc.yml
```diff
@@ -286,6 +289,8 @@ items:
           href: develop/get-javascript-intellisense-in-visual-studio.md
         - name: Convert JavaScript projects to TypeScript
           href: develop/convert-javascript-to-typescript.md
+      - name: Copy Script Lab snippet to Yo Office
+        href: overview/create-an-office-add-in-from-script-lab.md
```
samantharamon, line 293: Consider adding terms to `displayName` to help with discoverability during TOC filtering especially among devs new to Office Add-ins. Some examples I can think of are "code" or "code sample".

samantharamon (approved):
Looks good! Added comments for your consideration.

docs/toc.yml
AlexJerabek: Good suggestion. I settled on 'code samples' because the autocomplete will work if someone types 'code sample'.

Can you add some tags to the appropriate Outlook articles? Not a rush, but I feel like you know which ones are the correct choices to tag.

PR #4293, opened 2023-11-30
[Outlook] Specify platforms in docs

samantharamon:
- Replaces 'rich client' and 'desktop client' (where appropriate) with platform-specific terms, such as Outlook on Windows.
- Adds 'Office desktop application' to the glossary.


docs/outlook/match-strings-in-an-item-as-well-known-entities.md
```diff
@@ -122,23 +119,23 @@ There are a few facts and limits you should be aware of if you use well-known en
...
 
-- You can extract well-known entities from the first 2,000 characters in the item body, but not beyond that limit. This size limit helps balance the need for functionality and performance, so that Exchange Server and Outlook are not bogged down by parsing and identifying instances of well-known entities in large messages and appointments. Note that this limit is independent of whether the add-in specifies an `ItemHasKnownEntity` rule. If the add-in does use such a rule, note also the rule processing limit in item 2 below for the Outlook rich clients.
+- You can extract well-known entities from the first 2,000 characters in the item body, but not beyond that limit. This size limit helps balance the need for functionality and performance, so that Exchange Server and Outlook are not bogged down by parsing and identifying instances of well-known entities in large messages and appointments. Note that this limit is independent of whether the add-in specifies an `ItemHasKnownEntity` rule. If the add-in does use such a rule, note also the rule processing limit for Outlook on Windows and on Mac later in this section.
```
ElizabethSamuel-MSFT, line 122:
```suggestion
- You can extract well-known entities from the first 2,000 characters in the item body, but not beyond that limit. This size limit helps balance the need for functionality and performance, so that Exchange Server and Outlook aren't bogged down by parsing and identifying instances of well-known entities in large messages and appointments. Note that this limit is independent of whether the add-in specifies an `ItemHasKnownEntity` rule. If the add-in does use such a rule, note also the rule processing limit for Outlook on Windows and on Mac later in this section.
```

ElizabethSamuel-MSFT (approved):
Thanks for making this improvement. Just a couple of nits.

PR #4299, opened 2023-12-05
[all hosts] (Quickstarts) Reduce links in quickstarts

AlexJerabek:
Based on reader feedback, this PR reduces the number of links in the early parts of the quickstarts. The goal is to keep readers focused on the objective and not encourage them to wander around the docs until they have a working add-in.

A couple notes:
- The SSO quickstart has largely remained untouched, since that's not necessarily an introductory experience.
- I left the download link for Node.js, since explaining the various OS options seemed like more hassle.


Rick-Kirkham (approved):

samantharamon (approved):
Looks good! Just a couple of minor suggestions to fix contractions.

PR #4331, opened 2023-12-20
[all hosts] (IA) Fix quick start link text on index page

AlexJerabek:
This PR fixes the mislabeled Word quick start link. It also changes the text to prioritize for scanning - the only difference between the links is the application, so that should go first.


samantharamon (approved):
Great catch!

PR #4396, opened 2024-02-06
[Word] (wildcard search) Escaped special character

ElizabethSamuel-MSFT:
Adding example based on verbatim feedback. Clarifies how to search with a wildcard for an escaped special character.

### DIFF
--- a/docs/word/search-option-guidance.md
+++ b/docs/word/search-option-guidance.md
@@ -1,7 +1,7 @@
 ---
 title: Use search options in your Word add-in to find text 
 description: Learn to use search options in your Word add-in.
-ms.date: 03/15/2023
+ms.date: 02/06/2024
 ms.localizationpriority: medium
 ---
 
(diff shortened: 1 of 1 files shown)


docs/word/search-option-guidance.md
```diff
@@ -222,6 +222,34 @@ await Word.run(async (context) => {
 });
 ```
 
+### Search using a wildcard for an escaped special character
+
+```js
+// Run a batch operation against the Word object model.
+await Word.run(async (context) => {
+
+    // Queue a command to search the document with a wildcard for an escaped opening curly brace.
```
AlexJerabek, line 231: I'm not sure what this is actually looking for. Could we have a line outside of the code explaining what the code distinction is? And maybe an example?

AlexJerabek (changes requested):

AlexJerabek (approved):

docs/word/search-option-guidance.md
ElizabethSamuel-MSFT: Added

PR #4428, opened 2024-02-20
[Outlook] (multi-select) Add unified manifest support

samantharamon:
To be merged during GA of Outlook add-ins with unified manifest.

### DIFF
--- a/docs/outlook/item-multi-select.md
+++ b/docs/outlook/item-multi-select.md
@@ -1,7 +1,7 @@
 ---
 title: Activate your Outlook add-in on multiple messages
 description: Learn how to activate your Outlook add-in when multiple messages are selected.
-ms.date: 02/09/2024
+ms.date: 02/22/2024
 ms.topic: how-to
 ms.localizationpriority: medium
 ---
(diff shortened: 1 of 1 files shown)


docs/outlook/item-multi-select.md
```diff
@@ -21,8 +21,76 @@ Complete the [Outlook quick start](../quickstarts/outlook-quickstart.md?tabs=yeo
...
+
+1. Delete the second object of the "extensions.runtimes" array, whose "id" is "CommandsRuntime".
+
+1. In the "extensions.ribbons.tabs.controls" array, delete the second object, whose "id" is "ActionButton".
+
+1. Save your changes.
+
 > [!NOTE]
-> The item multi-select feature isn't currently supported in the [Unified manifest for Microsoft 365 (preview)](../develop/unified-manifest-overview.md), but the team is working on making this available.
+> Item multi-select can also be enabled without setting the "runtimes.actions.multiselect" property to `true` if the "runtimes.actions.supportsNoItemContext" property is set to `true` in the manifest. To learn more, see [Activate your Outlook add-in without the Reading Pane enabled or a message selected](contextless.md).
```
Rick-Kirkham, line 91: This note is correct, but I wonder if doing this would be a good practice. Since the default for "multiselect" is false, normally the absence of it would imply that it is false. But if users follow the suggestion in this note, they'll have a manifest where "multiselect" is not present but is true. I think this would be misleading and potentially confusing to a developer who is looking at the manifest to troubleshoot some problem. If I'd known what the product team was planning with this property, I'd have argued against this design. But I guess that ship has sailed, so I'd recommend that we just not tell people about this **implicit** setting of "multiselect" to true. If they always explicitly include it and set it to true when it should be, then their manifest is self-documenting.

Rick-Kirkham (approved):
Approving but with one major concern.

docs/outlook/item-multi-select.md
samantharamon: That's a great point. Thanks for bringing it up. I removed this note.

PR #4487, opened 2024-03-28
[Outlook] (debug) UseDirectDebugger should be a DWORD value not a registry key

Flyingdot:
Adding `UseDirectDebugger` as a registry key does not work. It needs to be a DWORD value.


ElizabethSamuel-MSFT (approved):
@Flyingdot Thanks for helping to improve the documentation!

PR #4598, opened 2024-05-31
[Outlook] (attachments) Document how to remove inline attachments

samantharamon:
Related PR: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69733


AlexJerabek (approved):
Looks good! There also might be an opportunity for a larger sample showing how to remove inline attachments in the future, but that's obviously beyond the scope of this PR.

PR #4624, opened 2024-06-18
[All hosts] (COM) Document how to obtain the ProgId of a COM add-in

samantharamon:
- Documents how to obtain the ProgId of a COM add-in.
- Removes VSTO references as these add-ins don't have a ProgId.


davidchesnut (approved):
Looks great!

PR #4627, opened 2024-06-19
[all hosts] (Testing) Update cache clear guidelines for Mac

alison-mk:


docs/testing/clear-cache.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Clear the Office cache
 description: Learn how to clear the Office cache on your computer.
-ms.date: 03/13/2023
```
Rick-Kirkham, line 4: I don't think you want to change the date of this file if you're not changing its contents.

Rick-Kirkham (approved):
Approving with suggestion.

docs/testing/clear-cache.md
alison-mk: I updated this file date because the include I did edit displays here, and I want to signal to customers that the cache clear instructions have gotten a substantial update recently (instructions for how to remove the manifest on Mac). Perhaps that's not the goal of the file dates?

Rick-Kirkham: OK. That makes sense.

PR #4631, opened 2024-06-21
[PowerPoint] (quick start) Add for content add-in

ElizabethSamuel-MSFT:


alison-mk (approved):
Both this and the default content add-in (with no changes) crash PowerPoint for me -- but that might be specific to my machine. Since it works for you, I think we can go ahead and publish.

PR #4683, opened 2024-07-22
[PowerPoint] (general) Removing note about limited PPT API support

AlexJerabek:
I think the PPT space has grown enough to remove this comment. Thoughts?


alison-mk (approved):
Agreed. Thanks!

PR #4700, opened 2024-08-02
[Outlook] (pinning) Clarify pinning behavior across Outlook modes

samantharamon:


docs/outlook/pinnable-taskpane.md
```diff
@@ -26,6 +26,10 @@ However, by default, if a user has an add-in task pane open for a message in the
 > - Appointments/Meetings
 > - Outlook.com
 
+## Supported Outlook modes
+
+Pinnable task panes are supported in both the Message Compose and Message Read modes in Outlook. However, pinning isn't supported across different modes. For example, if a user pins the task pane of an add-in while reading a message, then creates a new message, they won't see the add-in's task pane from the message they're composing. To view the task pane, the user must activate the add-in from the message they're composing. If the user then pins the task pane, the task pane remains pinned the next time the user composes another message.
```
exextoc, line 31: I think this is good. 

I'm wondering if we should include something like: "This is because A MessageCompose extension point could have completely different buttons/taskpanes than A MessageRead extension point" 

or if you think that's sort of like "apologizing language" or "making excuses" and we shouldn't include that in our documents?

exextoc (approved):
Made a small suggestion for you to consider...but I'm also fine with what you have. 

(timwan)

docs/outlook/pinnable-taskpane.md
samantharamon: Thanks for the feedback. Included this in the update.

ElizabethSamuel-MSFT (approved):

PR #4701, opened 2024-08-02
[all hosts] (Launch) Add article about custom protocol handlers

AlexJerabek:
This PR describes how to setup custom protocol handlers that launch add-ins.

### DIFF
--- a/docs/publish/custom-protocol-handler.md
+++ b/docs/publish/custom-protocol-handler.md
@@ -8,7 +8,7 @@ ms.localizationpriority: medium
 
 # Trust custom protocol handlers to launch add-ins
 
-Protocol handlers are registered with the operating system to allow an app to be launched from a URI (for example, how `mailto:` launchs an email client). Add-ins can also be launched from protocol handlers. This article explains how to automatically trust these custom protocol handlers by using admin group policies.
+Protocol handlers are registered with the operating system to allow an app to be launched from a URI (for example, how `mailto:` launches an email client). Add-ins can also be launched from protocol handlers. This article explains how to automatically trust these custom protocol handlers by using admin group policies.
 
 Every add-in and protocol pair needs to be trusted. This either comes in the form end-user consent or admin group policies. Similarly, admins can block certain add-in and protocol pairs.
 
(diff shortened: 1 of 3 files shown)


emmasab:
We may want to update the intro to make it clearer this is not required to use protocol handlers with an add-in. This is specifically for IT admins to set trusted / blocked protocols via group policy. If the reg key is not set, the end-user still will get a prompt to allow the protocol to execute. We may want to use "trust" a custom protocol handler instead of "enable" and "register"

I wonder if there is a better spot in the table of content for this doc as this is not related to add-in "development" per se, this is closer to add-in deployment/administration by enterprise IT.

croicu (approved):

docs/publish/custom-protocol-handler.md
```diff
@@ -0,0 +1,105 @@
...
+Protocol handlers are registered with the operating system to allow an app to be launched from a URI (for example, how `mailto:` launchs an email client). Add-ins can also be launched from protocol handlers. This article explains how to automatically trust these custom protocol handlers by using admin group policies.
+
+Every add-in and protocol pair needs to be trusted. This either comes in the form end-user consent or admin group policies. Similarly, admins can block certain add-in and protocol pairs.
+
+> [!IMPORTANT]
+> This information about trusting custom protocol handlers that launch add-ins through the registry only pertains to Windows.
+
+## Registry key format
+
+To automatically trust a custom protocol handler that launches an add-in, create a registry key at one of the following locations. Note that `<add-in id>` refers to the [Id element](/javascript/api/manifest/id) specified in your add-in's manifest.
```
Rick-Kirkham, line 20: If this feature works with add-ins using the unified manifest, then we should mention where the ID is found in that manifest.

```suggestion
To automatically trust a custom protocol handler that launches an add-in, create a registry key at one of the following locations. Note that `<add-in id>` refers to the [Id element](/javascript/api/manifest/id) specified in the add-in only manifest or to the "id" property of the unified manifest..
```

Rick-Kirkham (approved):
Approving with a comment.

PR #4278, opened 2023-11-15
[All hosts] (localStorage) Document recent update to Chromium-based browsers

samantharamon:
Fixes #4249.

PR #4702, opened 2024-08-05
[all hosts] (Browsers) Update based on latest information about Google Chrome 3P cookies

AlexJerabek:
These updates are based on the following two articles:
- https://www.reuters.com/technology/google-scraps-plan-remove-cookies-chrome-2024-07-22/
- https://developers.google.com/privacy-sandbox/3pcd/temporary-exceptions/storage-partitioning-deprecation-trial

In summary, Google is pivoting their 3P cookie strategy to not explicitly block them anymore. Additionally, the flag to work around storage partitioning in Chrome is being removed in the latest builds.


samantharamon (changes requested):
Thanks for catching this change. The updates look good, but you may also need to update additional articles that mention the flag workaround (see https://github.com/OfficeDev/office-js-docs-pr/pull/4278).

AlexJerabek:
Good catch @samantharamon! Is that line about Outlook for the web not loading snippets still true?

samantharamon (approved):

samantharamon:
> Good catch @samantharamon! Is that line about Outlook for the web not loading snippets still true?

Yes.

PR #4774, opened 2024-09-05
[all hosts] Remove extra param from convert command

ElizabethSamuel-MSFT:
I tried `npx office-addin-manifest-converter convert -m <relative-path-to-XML-manifest>` but I got an error message "error: unknown option '-m'". When I checked the [npm package docs](https://www.npmjs.com/package/office-addin-manifest-converter), there's no '-m'. I tried without '-m' then the command succeeded.


Rick-Kirkham (approved):
IMHO, a purely technical fix that you've checked with the actual code like this doesn't really need a review.

PR #4826, opened 2024-09-30
[All hosts] (glossary) Define function commands

samantharamon:


AlexJerabek (approved):
Thanks for adding this!

PR #4845, opened 2024-10-10
[Word] (troubleshooting) Add more entries

ElizabethSamuel-MSFT:


docs/word/word-add-ins-troubleshooting.md
```diff
@@ -10,12 +10,22 @@ ms.localizationpriority: medium
 
 This article discusses troubleshooting issues that are unique to Word. Use the feedback tool at the end of the page to suggest other issues that can be added to this article.
 
+## All selected ranges aren't recognized
+
+If non-contiguous or discontinuous selections are made, the Word API only operates on the last contiguous range in the selection. An unexpected case of this is when you select a column in a table then call, for example, [Document.getSelection](/javascript/api/word/word.document#word-word-document-getselection-member(1)), only the final cell in the selection is returned by the API. Although the selection of a column seems continuous, the API recognizes it as a non-contiguous selection (e.g., a cell per row).
```
AlexJerabek, line 15: What's the difference between "non-contiguous" and "discontinuous"?
```suggestion
If non-contiguous or discontinuous selections are made, the Word API only operates on the last contiguous range in the selection. An unexpected case of this is when you select a column in a table then call, for example, [Document.getSelection](/javascript/api/word/word.document#word-word-document-getselection-member(1)), only the final cell in the selection is returned by the API. Although the selection of a column seems contiguous, the API recognizes it as a non-contiguous selection (e.g., a cell per row).
```

AlexJerabek (approved):
Approving with a suggestion and question.

docs/word/word-add-ins-troubleshooting.md
ElizabethSamuel-MSFT: they have similar meanings for this context - non-contiguous means not adjoining or next to, discontinuous means that there are gaps. I'll just use contiguous/noncontiguous for this section

PR #4852, opened 2024-10-14
[Excel] (Custom Functions) Add new Get started article

alison-mk:


grangeryy (approved):

grangeryy:
Thank you @alison-mk  for putting things together and I've reviewed and approved it!

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Nesting in custom functions
+
+A custom function can accept another custom function as an argument, ensuring that any dependent values are updated during recalculation. The recalculation of the outer custom function depends on the result of the nested function, leading to increased time consumption with each additional nested function. Minimize the number of nested levels in your custom functions to improve recalculation efficiency. The following code snippets demonstrate two approaches that produce similar outputs. **Option 1** is more likely to reduce the nested levels for end users when adding values in the workbook compared to **Option 2**.
+
+#### Option 1: Less nesting
+
+```js
+    /**
```
davidchesnut, line 37: How does this nest another custom function? I can call this with Add(1,2,3) (no custom functions).

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,112 @@
...
+      let total = 0;
+      args.forEach(value => {
+        total += value;
+      });
+     
+      return total;
+    }
+```
+
+#### Option 2: More nesting
```
davidchesnut, line 51: If this is the "bad" example, might be better to start with this, explain why it causes extra calc, and then show the better example. right now it looks like you can use either option equally well.

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,112 @@
...
+    * @customfunction
+    */
+    function Add(arg1: number, arg2: number): number {
+      return arg1 + arg2;
+    }
+```
+
+### Excel calculation mode
+
+Excel has three calculation modes: Automatic, Automatic Except Tables, and Manual. For a description of these calculation modes, see [Excel Recalculation](/office/client-developer/excel/excel-recalculation). The most frequently used calculation mode for custom functions add-ins is manual calculation mode. Set the calculation mode for your add-in with the [Excel.CalculationMode enum](/javascript/api/excel/excel.calculationmode) based on your scenario. Note that automatic calculation mode may trigger recalculation often and reduce the efficiency of your add-in.
```
davidchesnut, line 65: Might make a clearer call to action here if the goal is to make your custom functions more efficient by reading the article. Perhaps the goal is to "Refer to the article and determine which calculation mode will work best for your custom function design..."

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,112 @@
...
+
+Custom functions add-ins allow for flexible designs, which means that different add-in designs can produce the same output for your end users.
+
+### Multiple results
+
+You can return multiple results from your custom function with multiple functions or with one function.
+
+To return multiple results with one function, use a dynamic array. This is usually the recommended approach because dynamic arrays only require updating a single cell to trigger recalculation for all results. To learn more about dynamic arrays in custom functions, see [Return multiple results from your custom function](custom-functions-dynamic-arrays.md).
+
+Another way to return multiple results is to use multiple functions and return a single result for each function. A benefit of using multiple functions is that your end user can decide precisely which formula they want to update and then only trigger recalculation for that formula.
```
davidchesnut, line 81: Might be good to show example like in the spec. Also point out the tradeoff of how one function for multiple results is less efficient across a larger dataset. Especially if relying on an external service.

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Complex data structures
```
davidchesnut, line 83: I'm unclear on how the subsections here related to complex data structures. They look like they apply to basic data types as well.

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,112 @@
...
+
+[Data types](custom-functions-data-types-concepts.md) are the best way to handle complex data structures in custom functions add-ins. Data types support [Excel errors](custom-functions-errors.md) and [formatted number values](custom-functions-data-types-concepts.md#output-a-formatted-number-value). Data types also allow for designing [entity value cards](excel-data-types-entity-card.md), extending Excel data beyond the 2-dimensional grid.
+
+## Improve the user experience of remote data calls
+
+Custom functions can fetch data from remote locations beyond the workbook, such as the web or a server. For more information about fetching remote data, see [Receive and handle data with custom functions](custom-functions-web-reqs.md). To maintain efficiency when making remote data calls, consider batching external calls, minimizing roundtrip duration for each call, and including messages in your add-in to communicate delays to your end user.
```
davidchesnut, line 89: Would be good to use the term "external service" to tie in with the graphic at the beginning and connect the dots here.

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,112 @@
...
+### Batch custom function remote calls
+
+If your custom functions call a remote service, use a batching pattern to reduce the number of network calls to the remote service. To learn more, see [Batching custom function calls for a remote service](custom-functions-batching.md).
+
+### Minimize roundtrip duration
+
+Remote service connections can have a large impact on custom function performance. To reduce this impact, use these strategies:
+
+- Server-heavy processing should be handled efficiently in the remote server to shorten the end-to-end latency for a custom function. For example, have parallel computing designed on the server. If your service is deployed on Azure, consider optimization using [high-performance computing on Azure](/azure/architecture/topics/high-performance-computing).
+- Reduce the number of service calls by optimizing the add-in flow. For example, only send necessary calls to a remote service.
```
davidchesnut, line 100: I would image we could have another section about using a cache for repeat CF calls, or common calls. Check with Phoebe.

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Improve user-perceived performance through add-in UX
+
+If a delay while calling a remote service is inevitable, consider providing messages through the add-in task pane to explain the delay to your end users. This information helps manage their expectations. The following image shows an example.
```
davidchesnut, line 104: I would mention that they will see the "busy" text in the cell while processing, so it's not like you should always think to use the task pane. But if you fell you must might be good to point out a pattern here of how it works. You need CF runtime, set the TP image on CF entry, then turn off the TP image on CF exit. Plus the task pane has to keep track if multiple CF long processes are running (which could get complicated).

docs/excel/custom-functions-get-started.md
alison-mk: @grangeryy Can you answer David's question here?

docs/excel/custom-functions-get-started.md
alison-mk: My thinking is that *all* Excel data types are a "complex data structure" from the perspective of custom functions, and the goal of this section is the communicate that data types offer a way for custom functions to handle complex data structures.

docs/excel/custom-functions-get-started.md
alison-mk: @grangeryy Can you take a look at David's suggestion here?

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,112 @@
+---
+description: Get started with Excel custom functions for Office Add-ins.
+title: Get started with custom functions in Excel
```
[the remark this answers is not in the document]
alison-mk: Do you recommend changing the title throughout (the H1 in the article, the TOC name), or only here?

docs/excel/custom-functions-get-started.md
alison-mk: Thanks for the feedback! I didn't end up restructuring, but I did add more explanation about which option is recommended and why.

docs/excel/custom-functions-get-started.md
alison-mk: I added a note about the BUSY error and also a link to the Share data and events between Excel custom functions and the task pane tutorial so customers can continue learning about the task pane if they want.

grangeryy (approved):

davidchesnut (approved):
Looks good! A few suggestions in the comments.

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,159 @@
...
+
+- Server-heavy processing should be handled efficiently in the remote server to shorten the end-to-end latency for a custom function. For example, have parallel computing designed on the server. If your service is deployed on Azure, consider optimization using [high-performance computing on Azure](/azure/architecture/topics/high-performance-computing).
+- Reduce the number of service calls by optimizing the add-in flow. For example, only send necessary calls to a remote service.
+
+### Improve user-perceived performance through add-in user experience (UX)
+
+While a custom function is calling an external service, the cell with the custom function displays the **#BUSY!** error. If a delay while calling an external service is inevitable, consider providing messages through the add-in task pane to explain the delay to your end users. This information helps manage their expectations. The following image shows an example.
+
+:::image type="content" source="../images/custom-functions-delay-example.png" alt-text="The delay message says 'It may take some time as we are getting the data ready for you'.":::
```
lindalu-MSFT, line 118: Please lower your screen resolution to something around 1920 x _n_ and reshoot this image. Be sure it is maximum 1200 pixels wide. This ensures PWDs who use magnification (reflow) can see it. Also, the contrast in the text box doesn't pass MAS requirements. The text should be dark gray to black.

image

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,159 @@
...
+      // ...
+      // ...
+      return answer;
+    }
+    ```
+
+## See also
+
+- [Receive and handle data with custom functions](custom-functions-web-reqs.md)
+- [Batching custom function calls for a remote service](custom-functions-batching.md)
```
lindalu-MSFT, line 158:
```suggestion
- [Batch custom function calls for a remote service](custom-functions-batching.md)
```

docs/excel/custom-functions-get-started.md
```diff
@@ -0,0 +1,159 @@
...
+- [Custom functions tutorial](../tutorials/excel-tutorial-create-custom-functions.md)
```
lindalu-MSFT, line 159:
```suggestion
- [Create custom functions in Excel tutorial](../tutorials/excel-tutorial-create-custom-functions.md)
```

lindalu-MSFT (changes requested):
Please see my note on the custom-functions-delay-image. There's a few other suggestions, too. ~linda

docs/excel/custom-functions-get-started.md
alison-mk: @lindalu-MSFT Thanks for the review! I don't have access to the conditions that created the original image, but I adjusted text contrast of the task pane message. Could you take another look?

lindalu-MSFT (approved):
Thank for fixing the image! It's definitely got enough contrast now.

docs/excel/custom-functions-get-started.md
lindalu-MSFT: Looks good!

PR #4868, opened 2024-10-22
[all hosts] (Visual Studio) Revise VS JS->TS steps

### LINKED ISSUE #4867: Convert an Office Add-in project in Visual Studio to TypeScript
Some comments on this documentation:

- In the section "Create an npm configuration file," step 5 does not appear to be correct. I believe it is meant to be select Tools-Options-Projects and Solutions-Web Package Management-Package Restore and set the npm Project Properties Default Settings both to TRUE.
- In the section "Update the Java scripts," there is not a line **Office.initialize = function (reason) {**.  I am guessing that the new line goes immediately after the **Office.onReady(() => {** but this is not clear. If there is another way to get at these settings, I could not find it.

At this point there are numerous errors in the Home.ts and FunctionFile.ts files. One is straightforward, in displaySelectedCells there is a const textValue that is later referred to as textvalue. This would be great to fix as it is just a bug in the Home.js file.

The second error of which there are many states:

Build:An async function or method in ES5 requires the 'Promise' constructor.  Make sure you have a declaration for the 'Promise' constructor or include 'ES2015' in your '--lib' option.

Some googling got me to add **"lib": [ "dom", "es5", "es2015.promise" ]** to the compiler options in the tsconfig.json file. I am not sure that this is the correct solution but it allowed the addin to build and run without error. 

Could you update the documentation and perhaps the home.js code to remove the errors.  

thanks,

AlexJerabek:
Fixes #4867.

Some of the steps in this article were out-of-date with UI and template changes. This PR should address that.

### DIFF
--- a/docs/develop/convert-javascript-to-typescript.md
+++ b/docs/develop/convert-javascript-to-typescript.md
@@ -2,7 +2,7 @@
 title: Convert an Office Add-in project in Visual Studio to TypeScript
 description: Learn how to convert an Office Add-in project in Visual Studio to use TypeScript.
 ms.topic: how-to
-ms.date: 08/06/2024
+ms.date: 10/22/2024
 ms.localizationpriority: medium
 ---
 
@@ -57,7 +57,11 @@ You can use the Office Add-in template in Visual Studio to create an add-in that
         "noEmitOnError": true,
         "removeComments": false,
         "sourceMap": true,
-        "target": "es5"
+        "target": "es5",
+        "lib": [ 
+          "es2015",
+          "dom"
+        ]
       },
       "exclude": [
         "node_modules",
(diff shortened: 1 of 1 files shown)


docs/develop/convert-javascript-to-typescript.md
```diff
@@ -90,27 +94,21 @@ You can use the Office Add-in template in Visual Studio to create an add-in that
...
+1. Open the npm project properties by going to **Tools** > **Options**, then **Projects and Solutions** > **Web Package Management** > **Package Restore**. Set both **Restore On Project Open** and **Restore On Save** to "True". Select **OK** to save the settings.
 
 ### Update the JavaScript files
 
 Change your JavaScript files (**.js**) to TypeScript files (**.ts**). Then, make the necessary changes for them to compile. This section walks through the default files in a new project.
 
 1. Find the **Home.js** file and rename it to **Home.ts**.
 
+    > [!WARNING]
+    > There is currently a issue in template's **Home.js** file: the reference on line 95 to `textvalue` should be changed to `textValue`.
```
ElizabethSamuel-MSFT, line 106:
```suggestion
    > There's currently an issue in the template's **Home.js** file: the reference on line 95 to `textvalue` should be changed to `textValue`.
```

ElizabethSamuel-MSFT (approved):

PR #4889, opened 2024-11-08
[all hosts] (DevX) Remove unnecessary Windows-specific runtime logging instructions

AlexJerabek:
The Windows-only instructions for runtime logging aren't needed, since the easier-to-use tool now works with all supported versions of npm.

### DIFF
--- a/docs/testing/runtime-logging.md
+++ b/docs/testing/runtime-logging.md
@@ -1,7 +1,7 @@
 ---
 title: Debug your add-in with runtime logging
 description: Learn how to use runtime logging to debug your add-in.
-ms.date: 05/05/2023
+ms.date: 11/08/2024
 ms.localizationpriority: medium
 ---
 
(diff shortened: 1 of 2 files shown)


docs/testing/runtime-logging.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Debug your add-in with runtime logging
```
Rick-Kirkham, line 2: Consider using title-casing as per https://review.learn.microsoft.com/en-us/help/platform/seo-meta-title?branch=main

Rick-Kirkham (approved):
Approving with a suggestion to think about.

docs/testing/runtime-logging.md
AlexJerabek: This seems like it should be adjusted across the board. I'll make an internal item to address that.

PR #4936, opened 2024-12-06
[Outlook](NAA) Add section about identity and your services

davidchesnut:
Documents a common question about moving from the old Exchange token scenario of validating the ID token to use it in services.

### DIFF
--- a/docs/outlook/faq-nested-app-auth-outlook-legacy-tokens.md
+++ b/docs/outlook/faq-nested-app-auth-outlook-legacy-tokens.md
@@ -165,7 +165,7 @@ If you submit an issue, please include the following information.
 - Version of msal-browser.
 - Logs from msal-browser.
 
-## Developer troubleshooting questions
+## Developer questions
 
 ### How do I get more debug information from MSAL and NAA?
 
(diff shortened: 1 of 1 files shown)


docs/outlook/faq-nested-app-auth-outlook-legacy-tokens.md
```diff
@@ -165,7 +165,7 @@ If you submit an issue, please include the following information.
 - Version of msal-browser.
 - Logs from msal-browser.
 
-## Developer troubleshooting questions
+## Developer questions
```
lindalu-MSFT, line 168:
```suggestion

```
I don't think you need this heading level.

docs/outlook/faq-nested-app-auth-outlook-legacy-tokens.md
```diff
@@ -165,7 +165,7 @@ If you submit an issue, please include the following information.
...
 
 ### How do I get more debug information from MSAL and NAA?
```
lindalu-MSFT, line 170:
```suggestion
## How do I get more debug information from MSAL and NAA?
```

lindalu-MSFT (commented):
A few suggestions to commit.

docs/outlook/faq-nested-app-auth-outlook-legacy-tokens.md
davidchesnut: It adds it to the top inline TOC which is handy to keep this separate from the admin section.

PR #4953, opened 2024-12-18
[Outlook] (event-based activation) Document limitation in new Outlook on Windows

samantharamon:


docs/outlook/autolaunch.md
```diff
@@ -75,6 +75,10 @@ In all supported Outlook clients, the user must remain on the current mail item
 
 In addition to item switching, an event-based add-in also ceases operation when the user sends the message or appointment they're composing.
 
+### Event-based add-ins in new Outlook on Windows
```
AlexJerabek, line 78: What's the standard way to reference "New Outlook"? [Here](https://learn.microsoft.com/en-us/office/dev/add-ins/outlook/one-outlook), we say "the new Outlook". Without the "the", it seems like we should capitalize "New".
```suggestion
### Event-based add-ins in the new Outlook on Windows
```

samantharamon: Thanks for catching this. If the term ends with "on Windows", "the" should be included. Otherwise, I agree that "New" should be capitalized.

docs/outlook/autolaunch.md
```diff
@@ -75,6 +75,10 @@ In all supported Outlook clients, the user must remain on the current mail item
 
 In addition to item switching, an event-based add-in also ceases operation when the user sends the message or appointment they're composing.
 
+### Event-based add-ins in the new Outlook on Windows
+
+In new Outlook on Windows, you must keep the main window of the client open for the add-in to process the mail item. If the main window is minimized, the add-in will pause or stop working.
```
samantharamon, line 80:
```suggestion
In the new Outlook on Windows, you must keep the main window of the client open for the add-in to process the mail item. If the main window is minimized, the add-in will pause or stop working.
```

AlexJerabek (approved):
Thanks for making that change!

PR #4985, opened 2025-01-06
[Outlook] (identity) Note about token behavior in February

davidchesnut:
Adds a note to let folks know that if you turn on the Exchange tokens, they will remain on even when we start turning them off in February.


lindalu-MSFT (approved):
Looks good to me!

PR #5007, opened 2025-01-27
[All Hosts] (DevX) GCC now has a special URL

Rick-Kirkham:
jonahkarpman has approved by email

### DIFF
--- a/docs/publish/government-cloud-guidance.md
+++ b/docs/publish/government-cloud-guidance.md
@@ -24,7 +24,7 @@ For all government cloud (i.e. Sovereign Cloud) environments, the public Microso
 
 [Centralized deployment](/microsoft-365/admin/manage/centralized-deployment-of-add-ins) of add-ins outside of the store is still supported.
 
-## GCC High (GCCH), US Department of Defense (DOD), or other sovereign managed clouds
+## GCC, GCC High (GCCH), US Department of Defense (DOD), or other sovereign managed clouds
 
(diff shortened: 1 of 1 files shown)


docs/publish/government-cloud-guidance.md
```diff
@@ -24,17 +24,13 @@ For all government cloud (i.e. Sovereign Cloud) environments, the public Microso
 
 [Centralized deployment](/microsoft-365/admin/manage/centralized-deployment-of-add-ins) of add-ins outside of the store is still supported.
```
ElizabethSamuel-MSFT, line 26: Can we define GCC somewhere, especially since the section that had spelt it out is being removed?

docs/publish/government-cloud-guidance.md
```diff
@@ -24,17 +24,13 @@ For all government cloud (i.e. Sovereign Cloud) environments, the public Microso
...
-
-## GCC High (GCCH), US Department of Defense (DOD), or other sovereign managed clouds
+## GCC, GCC High (GCCH), US Department of Defense (DOD), or other sovereign managed clouds
 
 These government clouds remain internet-connected, but the set of public endpoints made available is severely restricted. One such restricted endpoint is the public endpoint for loading the Office JavaScript API library. The public CDN location for Office.js will not be accessible from within these environments. However, there will be an internal, per-cloud Microsoft Office CDN provisioned with the same resource. This means the endpoint URL to access Office.js will be different and your Office Add-in may need some level of customization to run. Given the additional restrictions, it's likely that any solution provided to customers will require hosting provider services within the environment as well, necessitating additional customizations. You'll need to determine the best way to provide your solution to customers, such that it conforms to the additional restrictions imposed on services running within the boundaries of these environments. The Office JavaScript Library CDN URLs are as follows:
```
ElizabethSamuel-MSFT, line 29:
```suggestion
These government clouds remain internet-connected, but the set of public endpoints made available is severely restricted. One such restricted endpoint is the public endpoint for loading the Office JavaScript API library. The public CDN location for Office.js won't be accessible from within these environments. However, there'll be an internal, per-cloud Microsoft Office CDN provisioned with the same resource. This means the endpoint URL to access Office.js will be different and your Office Add-in may need some level of customization to run. Given the additional restrictions, it's likely that any solution provided to customers will require hosting provider services within the environment as well, necessitating additional customizations. You'll need to determine the best way to provide your solution to customers, such that it conforms to the additional restrictions imposed on services running within the boundaries of these environments. The Office JavaScript Library CDN URLs are as follows:
```

PR #5011, opened 2025-01-29
[Outlook] (equivalent add-ins) Clarify when to configure the Group Policy setting

samantharamon:


timwan10:
This may be saying the same thing except backwards... but do you think there should be a line that is very specific here, like "If the manifest contains an EquivalentAddin section, then adjusting the GPO is not necessary"

docs/develop/make-office-add-in-compatible-with-existing-com-add-in.md
```diff
@@ -28,7 +28,7 @@ Before you can specify an equivalent COM add-in, you must first identify its `Pr
 ### Configure the manifest
 
 > [!IMPORTANT]
-> Applies to Excel, Outlook, PowerPoint, and Word.
+> Applies to **Excel**, **Outlook**, **PowerPoint**, and **Word**.
```
ElizabethSamuel-MSFT, line 31: Curious why we're bolding the applications?

samantharamon: It was based on a suggestion to make supported hosts mentioned inline more visible. For an example, see [Create custom contextual tabs in Office Add-ins](https://learn.microsoft.com/office/dev/add-ins/design/contextual-tabs#prerequisites).

ElizabethSamuel-MSFT: I see. I guess because it's already in a very short note, it doesn't seem like bolding would add to the visibility

samantharamon: That makes sense. I reverted the bold format.

PR #5114, opened 2025-04-09
[Excel] (Range) Add article about checkboxes

alison-mk:
This PR adds an article about checkboxes, based on the new checkboxes snippet: https://github.com/OfficeDev/office-js-snippets/pull/973


docs/excel/excel-add-ins-ranges-checkboxes.md
```diff
@@ -0,0 +1,83 @@
...
+    range.control = {
+      type: Excel.CellControlType.checkbox
+    };
+    await context.sync();
+});
+```
+
+## Change the value of a checkbox
+
+To check or uncheck a checkbox with the Excel JavaScript API, change the Boolean value in that cell. Use `Range.values` to change the value of a cell. The following code sample shows how to set the value of a cell to **TRUE**. Note that if the cell doesn't already display a checkbox, then the code sample simply changes the Boolean value of the cell.
```
davidchesnut, line 43:
```suggestion
To select or clear a checkbox with the Excel JavaScript API, change the Boolean value in that cell. Use `Range.values` to change the value of a cell. The following code sample shows how to set the value of a cell to **TRUE**. Note that if the cell doesn't already display a checkbox, then the code sample simply changes the Boolean value of the cell.
```

davidchesnut (approved):
Looks good, with a suggestion. Also please remove the image.png duplicate file.

lindalu-MSFT (approved):
Looks great!

PR #5129, opened 2025-04-16
[Excel](custom functions) Support formula value preview

davidchesnut:
New article on how to support formula value preview in your custom functions.


lindalu-MSFT (approved):
Looks good!

docs/excel/custom-functions-formula-value-preview.md
```diff
@@ -0,0 +1,46 @@
...
+
+By default, custom functions (for example `=getHousePrice(A1)`) can be previewed by the user. However, the following list shows some scenarios in which you may want to control how your custom function participates in formula value preview mode.
+
+- Your custom function calls one or more APIs that charge a rate for using them.
+- Your custom function accesses one or more scarce resources such as databases.
+- Your custom function takes significant time to calculate the result, and it wouldn’t be useful for a user during preview purposes.
+
+You can change the behavior of your custom function to return a mock value instead. To do this use the `invocation.isInValuePreview` property. The following code sample shows an example custom function named `getHousePrice` that looks up house prices through a monetized API. If `isInValuePreview` is `true`, the custom function returns a mock number to be used and avoids incurring any cost. If `isInValuePreview` is `false`, the custom function calls the API and returns the actual house price value for use in the Excel spreadsheet.
```
alison-mk, line 20: I just worked with eng to add the `isInValuePreview` flag to the type definitions on DefinitelyTyped—this part still isn't super clear to me, but my understanding is that `isInValuePreview` can only be read by an add-in. It can't be changed by an add-in. Could we clarify that here? Is it read-only? How can the value be changed?

alison-mk (commented):
I have two broad questions before approving, once they're resolved I can approve!

docs/excel/custom-functions-formula-value-preview.md
davidchesnut: I added read-only.

PR #5153, opened 2025-05-02
[PowerPoint] (tables) Change way to update values in existing table

ElizabethSamuel-MSFT:
In the current version of the [Table object](https://learn.microsoft.com/javascript/api/powerpoint/powerpoint.table), the properties and methods seem to be read-only / gets, including `values`.


davidchesnut (approved):
Looks good!

PR #5158, opened 2025-05-05
Replace en dashes with hyphens in command lines

### LINKED ISSUE #4996: Copy/Paste issue with Commands
If you copy and paste the commands in the article directly into Powershell, it leaves off the dash before "BlockLegacyExchangeTokens" and before "AllowLegacyExchangeTokens".  
Looks like the text is a long dash when it should be a short dash.  Had to open a microsoft case to realize this.

Image

AlexJerabek:
Fixes #4996.

Commands pasted into the command line should use the standard dash character, not the en dash (that text editors would replace the hyphen with).


davidchesnut (approved):
Looks good!

PR #5167, opened 2025-05-12
[All Hosts] (devx) Teams Toolkit becoming Agents Toolkit

Rick-Kirkham:
Changes from Teams Toolkit to "Microsoft 365 Agents Toolkit" (1st mention) or "Agents Toolkit" (subsequent mentions). 
Also, the name of the CLI is changing and the signature command of the CLI is changing from "teamsapp" to "atk".

DO NOT MERGE TO MAIN UNTIL AFTERNOON OF MAY 16TH.


AlexJerabek (approved):
Approving, but you should consider renaming the "teams-toolkit-overview" file. Keywords in the URL impact SEO. If you do that, you'll also need to add a redirect.

PR #5194, opened 2025-05-22
[Outlook] (legacy tokens) information on how the report works

davidchesnut:
Adds information on how the telemetry reporting works with dates and updates.

### DIFF
--- a/docs/outlook/turn-exchange-tokens-on-off.md
+++ b/docs/outlook/turn-exchange-tokens-on-off.md
@@ -79,7 +79,7 @@ Blocked:
 ]
 ```
 
-The report only shows a single entry per add-in. If multiple calls from many users are made from a single add-in for Exchange tokens, those calls appear in the report as one request. The date updates every seven days. In the previous example, the report shows Script lab being granted tokens on May 16. The date won't change unless Script lab continues to make token requests on May 23rd at which point the report will update the date.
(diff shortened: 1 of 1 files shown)


docs/outlook/turn-exchange-tokens-on-off.md
```diff
@@ -64,9 +64,26 @@ Blocked: []
...
+[
+        { "49d3b812-abda-45b9-b478-9bc464ce5b9c" : "2025-05-16" }
+]
+Blocked:
+[
+        { "49d3b812-abda-45b9-b478-9bc464ce5b9c" : "2025-02-25" }
+]
+```
+
+The report only shows a single entry per add-in. If multiple calls from many users are made from a single add-in for Exchange tokens, those calls appear in the report as one request. The date updates every seven days. In the previous example, the report shows Script lab being granted tokens on May 16. The date won't change unless Script lab continues to make token requests on May 23rd at which point the report will update the date.
```
lindalu-MSFT, line 82:
```suggestion
The report only shows a single entry per add-in. If multiple calls from many users are made from a single add-in for Exchange tokens, those calls appear in the report as one request. The date updates every seven days. In the previous example, the report shows Script lab being granted tokens on May 16. The date won't change unless Script Lab continues to make token requests on May 23rd at which point the report will update the date.
```

lindalu-MSFT (approved):

PR #5226, opened 2025-06-11
[Excel] (custom functions) unified manifest support for xll preference

Rick-Kirkham:
Cannot merge until the xllCustomFunctions property in the unified manifest is moved to its proper place as a child of the "prefer" property.
UPDATE: This has now happened.


Rick-Kirkham:
> Approving, but I can't confirm that the xllCustomFunctions property has been moved. It looks to me like it's still a child of `alternates` and not a child of `prefer`: https://learn.microsoft.com/en-us/microsoft-365/extensibility/schema/extension-xll-custom-functions?view=m365-app-prev&tabs=syntax
> 
> > Cannot merge until the xllCustomFunctions property in the unified manifest is moved to its proper place as a child of the "prefer" property.
> > UPDATE: This has now happened.

The fix isn't in the reference documentation, because it was just made last week and the reference hasn't been autogenerated since then. But the fix is in the actual schema: [DevPreview/MicrosoftTeams.schema.json](https://developer.microsoft.com/json-schemas/teams/vDevPreview/MicrosoftTeams.schema.json)

PR #5248, opened 2025-06-24
[all hosts] (TOC) Move copy snippet to yo link to different TOC section

ElizabethSamuel-MSFT:
Moves the "Copy Script Lab snippet to Yo Office" link to a different section in the TOC, to be loosely grouped with the "Explore Office JS APIs using Script Lab" link.


AlexJerabek (approved):
Looks good! Let's check back on the page view and satisfaction data in a couple months, but I think this flow is more logical.

PR #5240, opened 2025-06-18
[Excel] (Custom functions) Add new JSDoc tags and metadata reference info

alison-mk:
The new custom enums article is added in a separate PR: https://github.com/OfficeDev/office-js-docs-pr/pull/5252

PR #5252, opened 2025-06-25
[Excel] (Custom functions) Add custom enums article

alison-mk:
Note that the related JSDoc and metadata reference articles are updated in separate PR: https://github.com/OfficeDev/office-js-docs-pr/pull/5240


docs/excel/custom-functions-custom-enums.md
```diff
@@ -0,0 +1,174 @@
...
+description: Create custom enums to use with your custom functions.
+ms.date: 06/24/2025
+ms.localizationpriority: medium
+---
+
+# Create custom enums for your custom functions
+
+:::image type="content" source="../images/custom-functions-custom-enum-autocomplete.png" alt-text="The members of a custom enum displaying in the Excel AutoComplete menu.":::
+
+With custom enums, the members in your enum display in the Excel AutoComplete menu for your custom function. The planets in the preceding screenshot are members of a custom enum. This article describes how to create custom enums and use them as parameters in your custom functions.
```
AlexJerabek: The member-centric framing sounded really techincal. I think people will get that an enum is inherently comprised of its members. 
```suggestion
Custom enums give users Excel AutoComplete options for your custom functions. The members in your enum show up in the formula bar as suggestions. The planets in the preceding screenshot are an example of a custom enum that provides a set list. This article describes how to create custom enums and use them as parameters in your custom functions.
```

docs/excel/custom-functions-custom-enums.md
```diff
@@ -0,0 +1,174 @@
...
+/** 
+ * A sample function that uses the custom enum as a parameter.
+ * @customfunction 
+ */ 
+function getPlanets(value: PLANETS): any { 
+  return value; 
+} 
+```
+
+## Edit enums in JSON metadata
```
AlexJerabek: Should this section be in the "Manually create JSON metadata for custom functions" article instead? I think that would help keep this article simple and help the common use case of people just using TypeScript.

docs/excel/custom-functions-custom-enums.md
```diff
@@ -0,0 +1,174 @@
...
+}
+```
+
+## Localize your custom enums
+
+Localizing custom enums is similar to [localizing custom functions](custom-functions-naming.md#localize-custom-functions). Create a new JSON metadata file for each language.
+
+Note that only the `name` and `tooltip` properties within an enum should be localized to the target language. The `value` property should remain unchanged to avoid the need for handling multiple languages within your function body.
+
+The following JSON snippet shows a localized `values` object for the planet Mercury.
```
AlexJerabek: Consider adding the language that's shown here.
```suggestion
The following JSON snippet shows the Chinese language localized `values` object for the planet Mercury.
```

AlexJerabek (approved):
Approving with a couple suggestions.

docs/excel/custom-functions-custom-enums.md
alison-mk: Thanks! Great rephrasing.

docs/excel/custom-functions-custom-enums.md
alison-mk: Ok, I'll move it. Thanks for the feedback!

PR #5256, opened 2025-06-25
[NAA] Update FAQ for new error message

davidchesnut:
Message returned by GCAT when a request for legacy Exchange Online token is denied has changed. This PR reflects the new message.


AlexJerabek:
Approving, though that broken link should get fixed.

PR #5287, opened 2025-07-16
Update Differences between Smart Alerts and the on-send feature

naokiohmi:
On-Send feature is supported on New Outlook for Windows.


naokiohmi (commented):
adjusted to (new and classic)

PR #5303, opened 2025-07-28
[Outlook] (unified manifest) Document unified manifest support in Outlook on Mac and on mobile

samantharamon:
Clarifies guidance for running add-ins that use the unified manifest on platforms that don't directly support it. Also links applicable Outlook articles to [Office Add-ins with the unified app manifest for Microsoft 365](https://learn.microsoft.com//office/dev/add-ins/develop/unified-manifest-overview#client-and-platform-support) and [Add-ins for Outlook on mobile devices](https://learn.microsoft.com/office/dev/add-ins/outlook/outlook-mobile-addins#support-for-add-ins-with-the-unified-manifest-for-microsoft-365).

### DIFF
--- a/docs/develop/unified-manifest-overview.md
+++ b/docs/develop/unified-manifest-overview.md
@@ -94,7 +94,7 @@ The following tables lists which Office platforms directly support add-ins that 
 | Office on mobile | Not directly supported |
 
 > [!NOTE]
-> Custom add-ins or line-of-business (LOB) add-ins that use the unified manifest can't currently be deployed in the Microsoft 365 Admin Center.
+> If you're deploying an add-in in the [Microsoft 365 Admin Center](../publish/publish.md) and require it to run on platforms that don't directly support the unified manifest, the add-in must be a published AppSource add-in. Custom add-ins or line-of-business (LOB) add-ins that use the unified manifest can't currently be deployed in the Microsoft 365 Admin Center.
 
 ## Sample unified manifest
 


docs/develop/unified-manifest-overview.md
```diff
@@ -94,7 +94,7 @@ The following tables lists which Office platforms directly support add-ins that
 | Office on mobile | Not directly supported |
 
 > [!NOTE]
-> If you're deploying an add-in in the [Microsoft 365 Admin Center](../publish/publish.md) and require it to run on platforms that don't directly support the unified manifest, the add-in must be a published AppSource add-in. Custom add-ins or line-of-business (LOB) add-ins that use the unified manifest can't currently be deployed in the Microsoft 365 Admin Center.
+> Custom add-ins or line-of-business (LOB) add-ins that use the unified manifest can't currently be deployed in the Microsoft 365 Admin Center.
```
Rick-Kirkham: This might oversimplify a bit. An admin can obtain an add-in from AppSource and deploy it to all users from the Admin Center. This works even for add-ins that use the unified manifest. The hidden XML manifest that is generated is used on clients that don't support the JSON manifest.

Rick-Kirkham (approved):
Approving with a suggestion.

docs/develop/unified-manifest-overview.md
samantharamon: You're right. Reverted this change.

PR #5310, opened 2025-08-04
[All hosts] (event-based activation) Note limitation of initialization methods on Windows clients

samantharamon:


samantharamon:
> Looks good! While you're making this change, could you also add small edit to this article: https://learn.microsoft.com/en-us/office/dev/add-ins/develop/event-based-activation#event-based-add-in-limitations-in-excel-powerpoint-word-and-classic-outlook-on-windows
> 
> Please change the line "We recommend adding any startup logic needed by event handlers, such as checking the user's Outlook version, to the event handlers instead." to "We recommend adding any startup logic needed by event handlers, such as checking the user's client version, to the event handlers instead." (change "Outlook" to "client" like you have in your version). Thanks!

Good catch! Updated.

PR #5329, opened 2025-08-27
[Outlook] (drag and drop) Correct enum

samantharamon:


ElizabethSamuel-MSFT (approved):
BTW, I think it's strange that the event name includes the word "Event"

samantharamon:
> BTW, I think it's strange that the event name includes the word "Event"

Agreed. The team is having discussions about the name, but we wanted to update the docs with the current value so that devs can test the feature without running into issues.

PR #5364, opened 2025-09-20
[Excel] (Custom functions) Update low engagement articles

alison-mk:
This PR updates Custom Functions articles flagged for low engagement, using suggestions from GitHub Copilot that have then been reviewed and revised by a human. It also adds a markdownlint.json file to stop the markdown linter from flagging common and intentional design patterns in our articles (such as the first H1). 

Note: The changes to Excel files in this PR are trivial. Excel articles are updated in a different PR.

Related PRs: 
#5378 
#5374

PR #5374, opened 2025-09-23
[Excel] Update low engagement articles

alison-mk:
This PR updates Excel articles flagged for low engagement, using suggestions from GitHub Copilot that have then been reviewed and revised by a human. 

Related PRs: 
#5364
#5378

### DIFF
--- a/docs/excel/excel-add-ins-data-validation.md
+++ b/docs/excel/excel-add-ins-data-validation.md
@@ -7,7 +7,7 @@ ms.localizationpriority: medium
 
 # Add data validation to Excel ranges
 
-Use the Excel JavaScript API to enforce data quality. Apply rules and rely on Excel’s validation UI for prompts and error alerts. This article shows how to define rule types, configure prompts and error alerts, and remove or adjust validation. If you need background on Excel’s built-in validation UI, review these end-user articles:
+Use the Excel JavaScript API to enforce data quality. Apply rules and rely on Excel’s validation UI for prompts and error alerts. This article shows how to define rule types, configure prompts and error alerts, and remove or adjust validation. If you need background on Excel’s built-in validation UI, review these articles.
 
 - [Apply data validation to cells](https://support.microsoft.com/office/29fecbcc-d1b9-42c1-9d76-eff3ce5f7249)
 - [More on data validation](https://support.microsoft.com/office/f38dee73-9900-4ca6-9301-8a5f6e1f0c4c)
@@ -15,7 +15,7 @@ Use the Excel JavaScript API to enforce data quality. Apply rules and rely on Ex
 
 ## Programmatic control of data validation
 
-The `Range.dataValidation` property, which takes a [DataValidation](/javascript/api/excel/excel.datavalidation) object, is the entry point for programmatic control of data validation in Excel. The object exposes five properties:
+The `Range.dataValidation` property, which takes a [DataValidation](/javascript/api/excel/excel.datavalidation) object, is the entry point for programmatic control of data validation in Excel. The object has five properties:
 
 - `rule` &#8212; Defines what constitutes valid data for the range. See [DataValidationRule](/javascript/api/excel/excel.datavalidationrule).
 - `errorAlert` &#8212; Specifies whether an error pops up if the user enters invalid data, and defines the alert text, title, and style such as `information`, `warning`, and `stop`. See [DataValidationErrorAlert](/javascript/api/excel/excel.datavalidationerroralert).
@@ -155,7 +155,7 @@ await Excel.run(async (context) => {
 
 ## Create validation error alerts
 
-Create an error alert to guide the user when invalid data is entered. The following is an example. Key points:
+Create an error alert to guide the user when invalid data is entered. The following example creates a basic alert. Key points:
 
 - The `style` property determines whether the user gets an informational alert, a warning, or a "stop" alert. Only `stop` actually prevents the user from adding invalid data. The pop-ups for `warning` and `information` have options that allow the user enter the invalid data anyway.
 - The `showAlert` property defaults to `true`. This means that Excel will pop-up a generic alert (of type `stop`) unless you create a custom alert which either sets `showAlert` to `false` or sets a custom message, title, and style. This code sets a custom message and title.
@@ -182,7 +182,7 @@ For more information, see [DataValidationErrorAlert](/javascript/api/excel/excel
 
 ## Create validation prompts
 
-Create an instructional prompt that appears when the user selects the cell. Example:
+Create an instructional prompt that appears when the user selects the cell. This example tells the user about the positive number validation before they enter data.
 
 ```js
 await Excel.run(async (context) => {
--- a/docs/excel/excel-add-ins-data-validation.md
+++ b/docs/excel/excel-add-ins-data-validation.md
@@ -1,13 +1,13 @@
 ---
 title: Add data validation to Excel ranges
-description: Learn how the Excel JavaScript APIs enable your add-in to add automatic data validation to tables, columns, rows, and other ranges in a workbook.
-ms.date: 02/16/2022
+description: Use the Excel JavaScript API to programmatically apply and manage data validation rules for ranges, tables, and structured inputs.
+ms.date: 09/19/2025
 ms.localizationpriority: medium
 ---
 
 # Add data validation to Excel ranges
 
-The Excel JavaScript Library provides APIs to enable your add-in to add automatic data validation to tables, columns, rows, and other ranges in a workbook. To understand the concepts and the terminology of data validation, please see the following articles about how users add data validation through the Excel UI.
+Use the Excel JavaScript API to enforce data quality. Apply rules and rely on Excel’s validation UI for prompts and error alerts. This article shows how to define rule types, configure prompts and error alerts, and remove or adjust validation. If you need background on Excel’s built-in validation UI, review these end-user articles:
 
 - [Apply data validation to cells](https://support.microsoft.com/office/29fecbcc-d1b9-42c1-9d76-eff3ce5f7249)
 - [More on data validation](https://support.microsoft.com/office/f38dee73-9900-4ca6-9301-8a5f6e1f0c4c)
@@ -15,10 +15,10 @@ The Excel JavaScript Library provides APIs to enable your add-in to add automati
 
 ## Programmatic control of data validation
 
-The `Range.dataValidation` property, which takes a [DataValidation](/javascript/api/excel/excel.datavalidation) object, is the entry point for programmatic control of data validation in Excel. There are five properties to the `DataValidation` object:
+The `Range.dataValidation` property, which takes a [DataValidation](/javascript/api/excel/excel.datavalidation) object, is the entry point for programmatic control of data validation in Excel. The object exposes five properties:
 
 - `rule` &#8212; Defines what constitutes valid data for the range. See [DataValidationRule](/javascript/api/excel/excel.datavalidationrule).
-- `errorAlert` &#8212; Specifies whether an error pops up if the user enters invalid data, and defines the alert text, title, and style; for example, `information`, `warning`, and `stop`. See [DataValidationErrorAlert](/javascript/api/excel/excel.datavalidationerroralert).
+- `errorAlert` &#8212; Specifies whether an error pops up if the user enters invalid data, and defines the alert text, title, and style such as `information`, `warning`, and `stop`. See [DataValidationErrorAlert](/javascript/api/excel/excel.datavalidationerroralert).
 - `prompt` &#8212; Specifies whether a prompt appears when the user hovers over the range and defines the prompt message. See [DataValidationPrompt](/javascript/api/excel/excel.datavalidationprompt).
 - `ignoreBlanks` &#8212; Specifies whether the data validation rule applies to blank cells in the range. Defaults to `true`.
 - `type` &#8212; A read-only identification of the validation type, such as WholeNumber, Date, TextLength, etc. It is set indirectly when you set the `rule` property.
@@ -26,7 +26,7 @@ The `Range.dataValidation` property, which takes a [DataValidation](/javascript/
 > [!NOTE]
 > Data validation added programmatically behaves just like manually added data validation. In particular, note that data validation is triggered only if the user directly enters a value into a cell or copies and pastes a cell from elsewhere in the workbook and chooses the **Values** paste option. If the user copies a cell and does a plain paste into a range with data validation, validation is not triggered.
 
-## Creating validation rules
+## Create validation rules
 
 To add data validation to a range, your code must set the `rule` property of the `DataValidation` object in `Range.dataValidation`. This takes a [DataValidationRule](/javascript/api/excel/excel.datavalidationrule) object which has seven optional properties. *No more than one of these properties may be present in any `DataValidationRule` object.* The property that you include determines the type of validation.
 
@@ -38,10 +38,10 @@ The first three `DataValidationRule` properties (i.e., validation rule types) ta
 - `decimal` &#8212; Requires a decimal number in addition to any other validation specified by the `BasicDataValidation` object.
 - `textLength` &#8212; Applies the validation details in the `BasicDataValidation` object to the *length* of the cell's value.
 
-Here is an example of creating a validation rule. Note the following about this code.
+The following example creates a validation rule. Key points:
 
-- The `operator` is the binary operator `greaterThan`. Whenever you use a binary operator, the value that the user tries to enter in the cell is the left-hand operand and the value specified in `formula1` is the right-hand operand. So this rule says that only whole numbers that are greater than 0 are valid.
-- The `formula1` is a hard-coded number. If you don't know at coding time what the value should be, you can also use an Excel formula (as a string) for the value. For example, "=A3" and "=SUM(A4,B5)" could also be values of `formula1`.
+- The `operator` is the binary operator `greaterThan`. Whenever you use a binary operator, the value that the user tries to enter in the cell is the left-hand operand and the value specified in `formula1` is the right-hand operand. This rule says that only whole numbers greater than 0 are valid.
+- The `formula1` is a hard-coded number. If you don't know at coding time what the value should be, you can also use an Excel formula as a string (such as "=A3" or "=SUM(A4,B5)").
 
 ```js
 await Excel.run(async (context) => {
@@ -59,9 +59,9 @@ await Excel.run(async (context) => {
 });
 ```
 
-See [BasicDataValidation](/javascript/api/excel/excel.basicdatavalidation) for a list of the other binary operators.
+See [BasicDataValidation](/javascript/api/excel/excel.basicdatavalidation) for other binary operators.
 
-There are also two ternary operators: `between` and `notBetween`. To use these, you must specify the optional `formula2` property. The `formula1` and `formula2` values are the bounding operands. The value that the user tries to enter in the cell is the third (evaluated) operand. The following is an example of using the "Between" operator.
+There are also two ternary operators: `between` and `notBetween`. To use these, specify the optional `formula2` property. The `formula1` and `formula2` values are the bounding operands. The value that the user tries to enter in the cell is the third (evaluated) operand. The following is an example of using the "Between" operator.
 
 ```js
 await Excel.run(async (context) => {
@@ -106,11 +106,11 @@ await Excel.run(async (context) => {
 
 ### List validation rule type
 
-Use the `list` property in the `DataValidationRule` object to specify that the only valid values are those from a finite list. The following is an example. Note the following about this code.
+Use the `list` property in the `DataValidationRule` object to constrain values to a finite set. The following code sample demonstrates. Key points:
 
 - It assumes that there is a worksheet named "Names" and that the values in the range "A1:A3" are names.
-- The `source` property specifies the list of valid values. The string argument refers to a range containing the names. You can also assign a comma-delimited list; for example: "Sue, Ricky, Liz".
-- The `inCellDropDown` property specifies whether a drop-down control will appear in the cell when the user selects it. If set to `true`, then the drop-down appears with the list of values from the `source`.
+- The `source` property specifies the list of valid values. The string argument refers to a range containing the names. You can also assign a comma-delimited list such as "Sue, Ricky, Liz".
+- The `inCellDropDown` property specifies whether a drop-down control appears in the cell when the user selects it. If `true`, the drop-down appears with the list of values from the `source`.
 
 ```js
 await Excel.run(async (context) => {
@@ -131,11 +131,12 @@ await Excel.run(async (context) => {
 
 ### Custom validation rule type
 
-Use the `custom` property in the `DataValidationRule` object to specify a custom validation formula. The following is an example. Note the following about this code.
+Use the `custom` property to specify a custom validation formula. The following is an example. Key points:
 
-- It assumes there is a two-column table with columns **Athlete Name** and **Comments** in the A and B columns of the worksheet.
-- To reduce verbosity in the **Comments** column, it makes data that includes the athlete's name invalid.
-- `SEARCH(A2,B2)` returns the starting position, in string in B2, of the string in A2. If A2 is not contained in B2, it does not return a number. `ISNUMBER()` returns a boolean. So the `formula` property says that valid data for the **Comment** column is data that does not include the string in the **Athlete Name** column.
+- It assumes a there is two-column table with columns **Athlete Name** and **Comments** in the A and B columns of the worksheet.
+- To reduce verbosity in the **Comments** column, the rule makes data that includes the athlete's name invalid.
+- `SEARCH(A2,B2)` returns the starting position in B2 of the string in A2. If A2 isn't contained in B2, it doesn't return a number. 
+- `ISNUMBER()` returns a boolean. So the `formula` property says that valid data for **Comments** is data that doesn't include the **Athlete Name** string.
 
 ```js
 await Excel.run(async (context) => {
@@ -154,7 +155,7 @@ await Excel.run(async (context) => {
 
 ## Create validation error alerts
 
-You can a create custom error alert that appears when a user tries to enter invalid data in a cell. The following is a simple example. Note the following about this code.
+Create an error alert to guide the user when invalid data is entered. The following is an example. Key points:
 
 - The `style` property determines whether the user gets an informational alert, a warning, or a "stop" alert. Only `stop` actually prevents the user from adding invalid data. The pop-ups for `warning` and `information` have options that allow the user enter the invalid data anyway.
 - The `showAlert` property defaults to `true`. This means that Excel will pop-up a generic alert (of type `stop`) unless you create a custom alert which either sets `showAlert` to `false` or sets a custom message, title, and style. This code sets a custom message and title.
@@ -181,7 +182,7 @@ For more information, see [DataValidationErrorAlert](/javascript/api/excel/excel
 
 ## Create validation prompts
 
-You can create an instructional prompt that appears when a user hovers over, or selects, a cell to which data validation has been applied. The following is an example.
+Create an instructional prompt that appears when the user selects the cell. Example:
 
 ```js
 await Excel.run(async (context) => {
@@ -207,13 +208,18 @@ For more information, see [DataValidationPrompt](/javascript/api/excel/excel.dat
 To remove data validation from a range, call the  [Range.dataValidation.clear()](/javascript/api/excel/excel.datavalidation#excel-excel-datavalidation-clear-member(1)) method.
 
 ```js
-myrange.dataValidation.clear()
+myrange.dataValidation.clear();
 ```
 
-It isn't necessary that the range you clear is exactly the same range as a range on which you added data validation. If it isn't, only the overlapping cells, if any, of the two ranges are cleared. 
+The range you clear doesn’t need to precisely match the range on which you added data validation. If the two ranges aren't an exact match, only overlapping cells are cleared.
 
 > [!NOTE]
 > Clearing data validation from a range will also clear any data validation that a user has added manually to the range.
+
+## Next steps
+
+- Combine validation with events: [Events](excel-add-ins-events.md).
+- Add conditional formatting for stronger visual cues.
 
 ## See also
 
--- a/docs/excel/excel-add-ins-data-validation.md
+++ b/docs/excel/excel-add-ins-data-validation.md
@@ -135,7 +135,7 @@ Use the `custom` property to specify a custom validation formula. The following 
 
 - It assumes a there is two-column table with columns **Athlete Name** and **Comments** in the A and B columns of the worksheet.
 - To reduce verbosity in the **Comments** column, the rule makes data that includes the athlete's name invalid.
-- `SEARCH(A2,B2)` returns the starting position in B2 of the string in A2. If A2 isn't contained in B2, it doesn't return a number. 
+- `SEARCH(A2,B2)` returns the starting position in B2 of the string in A2. If A2 isn't contained in B2, it doesn't return a number.
 - `ISNUMBER()` returns a boolean. So the `formula` property says that valid data for **Comments** is data that doesn't include the **Athlete Name** string.
 
 ```js
@@ -219,7 +219,7 @@ The range you clear doesn’t need to precisely match the range on which you add
 ## Next steps
 
 - Combine validation with events: [Events](excel-add-ins-events.md).
-- Add conditional formatting for stronger visual cues.
+- Add [conditional formatting](excel-add-ins-conditional-formatting.md) for stronger visual cues.
 
 ## See also
 
--- a/docs/excel/excel-add-ins-delay-in-cell-edit.md
+++ b/docs/excel/excel-add-ins-delay-in-cell-edit.md
@@ -1,17 +1,42 @@
 ---
 title: Delay execution while cell is being edited
-description: Learn how to delay the execution of the Excel.run function when a cell is being edited.
-ms.date: 02/16/2022
+description: Defer an Excel.run batch until the user leaves cell edit mode instead of failing with an error.
+ms.date: 09/19/2025
 ms.localizationpriority: medium
 ---
 
-
 # Delay execution while cell is being edited
 
-`Excel.run` has an overload that takes in a [Excel.RunOptions](/javascript/api/excel/excel.runoptions) object. This contains a set of properties that affect platform behavior when the function runs. The following property is currently supported.
+When a user is actively editing a cell, some operations your add-in starts may fail immediately. Use the `delayForCellEdit` option to queue the batch until the user exits cell edit mode instead of throwing an error.
 
-- `delayForCellEdit`: Determines whether Excel delays the batch request until the user exits cell edit mode. When `true`, the batch request is delayed and runs when the user exits cell edit mode. When `false`, the batch request automatically fails if the user is in cell edit mode (causing an error to reach the user). The default behavior with no `delayForCellEdit` property specified is equivalent to when it is `false`.
+`Excel.run` has an overload that takes an [Excel.RunOptions](/javascript/api/excel/excel.runoptions) object. It supports the following property relevant to this scenario:
+
+- `delayForCellEdit`: When `true`, Excel queues the batch until the user exits cell edit mode. When `false`, the batch fails immediately if the user is editing. If omitted, the default value is `false`.
+
+## Behavior comparison
+
+| User editing? | delayForCellEdit = false (default) | delayForCellEdit = true |
+|---------------|------------------------------------|--------------------------|
+| No | Batch runs immediately | Batch runs immediately |
+| Yes | Batch fails (`InvalidOperation` error) | Batch waits; then runs after edit is committed or canceled |
+
+## Example
 
 ```js
-await Excel.run({ delayForCellEdit: true }, async (context) => { ... });
+await Excel.run({ delayForCellEdit: true }, async (context) => {
+  const sheet = context.workbook.worksheets.getActiveWorksheet();
+  const range = sheet.getRange("A1");
+  range.values = [["Updated while user was editing elsewhere"]];
+  await context.sync();
+});
 ```
+
+## Guidance
+
+- Use `delayForCellEdit` only when user-initiated commands may overlap active cell editing, such as with a ribbon button that triggers a bulk update.
+- Consider showing a status indicator in your add-in if queued work may be lengthy.
+- Avoid chaining multiple long-running delayed batches. Queue depth can create a perceived lag for your add-in users.
+
+## Next steps
+
+Explore related user-state strategies in [Events](excel-add-ins-events.md), such as reacting to selection changes, and combine with `delayForCellEdit` to improve your add-in's robustness.
--- a/docs/excel/excel-add-ins-delay-in-cell-edit.md
+++ b/docs/excel/excel-add-ins-delay-in-cell-edit.md
@@ -7,15 +7,15 @@ ms.localizationpriority: medium
 
 # Delay execution while cell is being edited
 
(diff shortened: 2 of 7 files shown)


docs/excel/excel-add-ins-data-validation.md
```diff
@@ -1,32 +1,32 @@
...
-ms.date: 02/16/2022
+description: Use the Excel JavaScript API to programmatically apply and manage data validation rules for ranges, tables, and structured inputs.
+ms.date: 09/19/2025
 ms.localizationpriority: medium
 ---
 
 # Add data validation to Excel ranges
 
-The Excel JavaScript Library provides APIs to enable your add-in to add automatic data validation to tables, columns, rows, and other ranges in a workbook. To understand the concepts and the terminology of data validation, please see the following articles about how users add data validation through the Excel UI.
+Use the Excel JavaScript API to enforce data quality. Apply rules and rely on Excel’s validation UI for prompts and error alerts. This article shows how to define rule types, configure prompts and error alerts, and remove or adjust validation. If you need background on Excel’s built-in validation UI, review these end-user articles:
```
AlexJerabek: Technically, this is a full sentence before the bulleted list, so we should use a period (unless that's changed recently). I also don't think it's necessary to call the audience of the linked docs.
```suggestion
Use the Excel JavaScript API to enforce data quality. Apply rules and rely on Excel’s validation UI for prompts and error alerts. This article shows how to define rule types, configure prompts and error alerts, and remove or adjust validation. If you need background on Excel’s built-in validation UI, review these articles.
```

docs/excel/excel-add-ins-data-validation.md
```diff
@@ -1,32 +1,32 @@
...
 
 - [Apply data validation to cells](https://support.microsoft.com/office/29fecbcc-d1b9-42c1-9d76-eff3ce5f7249)
 - [More on data validation](https://support.microsoft.com/office/f38dee73-9900-4ca6-9301-8a5f6e1f0c4c)
 - [Description and examples of data validation in Excel](https://support.microsoft.com/help/211485)
 
 ## Programmatic control of data validation
 
-The `Range.dataValidation` property, which takes a [DataValidation](/javascript/api/excel/excel.datavalidation) object, is the entry point for programmatic control of data validation in Excel. There are five properties to the `DataValidation` object:
+The `Range.dataValidation` property, which takes a [DataValidation](/javascript/api/excel/excel.datavalidation) object, is the entry point for programmatic control of data validation in Excel. The object exposes five properties:
```
AlexJerabek:
```suggestion
The `Range.dataValidation` property, which takes a [DataValidation](/javascript/api/excel/excel.datavalidation) object, is the entry point for programmatic control of data validation in Excel. The object has five properties:
```

docs/excel/excel-add-ins-data-validation.md
```diff
@@ -38,10 +38,10 @@ The first three `DataValidationRule` properties (i.e., validation rule types) ta
 - `decimal` &#8212; Requires a decimal number in addition to any other validation specified by the `BasicDataValidation` object.
 - `textLength` &#8212; Applies the validation details in the `BasicDataValidation` object to the *length* of the cell's value.
 
-Here is an example of creating a validation rule. Note the following about this code.
+The following example creates a validation rule. Key points:
```
AlexJerabek, line 41: I wonder if instead of "Key points" the phrase "Please note" would be more consistent with our existing tone and style.

docs/excel/excel-add-ins-data-validation.md
```diff
@@ -154,7 +155,7 @@ await Excel.run(async (context) => {
 
 ## Create validation error alerts
 
-You can a create custom error alert that appears when a user tries to enter invalid data in a cell. The following is a simple example. Note the following about this code.
+Create an error alert to guide the user when invalid data is entered. The following is an example. Key points:
```
AlexJerabek:
```suggestion
Create an error alert to guide the user when invalid data is entered. The following example creates a basic alert. Key points:
```

docs/excel/excel-add-ins-data-validation.md
```diff
@@ -181,7 +182,7 @@ For more information, see [DataValidationErrorAlert](/javascript/api/excel/excel
 
 ## Create validation prompts
 
-You can create an instructional prompt that appears when a user hovers over, or selects, a cell to which data validation has been applied. The following is an example.
+Create an instructional prompt that appears when the user selects the cell. Example:
```
AlexJerabek:
```suggestion
Create an instructional prompt that appears when the user selects the cell. This example tells the user about the positive number validation before they enter data.
```

docs/excel/excel-add-ins-data-validation.md
```diff
@@ -207,16 +208,21 @@ For more information, see [DataValidationPrompt](/javascript/api/excel/excel.dat
...
-It isn't necessary that the range you clear is exactly the same range as a range on which you added data validation. If it isn't, only the overlapping cells, if any, of the two ranges are cleared. 
+The range you clear doesn’t need to precisely match the range on which you added data validation. If the two ranges aren't an exact match, only overlapping cells are cleared.
 
 > [!NOTE]
 > Clearing data validation from a range will also clear any data validation that a user has added manually to the range.
 
+## Next steps
+
+- Combine validation with events: [Events](excel-add-ins-events.md).
+- Add conditional formatting for stronger visual cues.
```
AlexJerabek: Link to conditional formatting?

docs/excel/excel-add-ins-data-validation.md
```diff
@@ -207,16 +208,21 @@ For more information, see [DataValidationPrompt](/javascript/api/excel/excel.dat
...
 
-It isn't necessary that the range you clear is exactly the same range as a range on which you added data validation. If it isn't, only the overlapping cells, if any, of the two ranges are cleared. 
+The range you clear doesn’t need to precisely match the range on which you added data validation. If the two ranges aren't an exact match, only overlapping cells are cleared.
 
 > [!NOTE]
 > Clearing data validation from a range will also clear any data validation that a user has added manually to the range.
 
+## Next steps
+
+- Combine validation with events: [Events](excel-add-ins-events.md).
```
AlexJerabek, line 221: This sounds good in theory, but I'm not sure what scenario this is actually hinting at.

docs/excel/excel-add-ins-delay-in-cell-edit.md
```diff
@@ -1,17 +1,42 @@
...
+description: Defer an Excel.run batch until the user leaves cell edit mode instead of failing with an error.
+ms.date: 09/19/2025
 ms.localizationpriority: medium
 ---
 
-
 # Delay execution while cell is being edited
 
-`Excel.run` has an overload that takes in a [Excel.RunOptions](/javascript/api/excel/excel.runoptions) object. This contains a set of properties that affect platform behavior when the function runs. The following property is currently supported.
+When a user is actively editing a cell, some operations your add-in starts may fail immediately. Use the `delayForCellEdit` option to queue the batch until the user exits cell edit mode instead of throwing an error.
```
AlexJerabek:
```suggestion
When a user is actively editing a cell, some add-in operations may fail immediately. Use the `delayForCellEdit` option to queue the batch until the user exits cell edit mode instead of throwing an error.
```

docs/excel/excel-add-ins-delay-in-cell-edit.md
```diff
@@ -1,17 +1,42 @@
...
+
+`Excel.run` has an overload that takes an [Excel.RunOptions](/javascript/api/excel/excel.runoptions) object. It supports the following property relevant to this scenario:
```
AlexJerabek:
```suggestion
`Excel.run` has an overload that takes an [Excel.RunOptions](/javascript/api/excel/excel.runoptions) object. It supports the following property relevant to this scenario.
```

docs/excel/excel-add-ins-delay-in-cell-edit.md
```diff
@@ -1,17 +1,42 @@
...
+
+- `delayForCellEdit`: When `true`, Excel queues the batch until the user exits cell edit mode. When `false`, the batch fails immediately if the user is editing. If omitted, the default value is `false`.
```
AlexJerabek: Omission is implied by need for a default.
```suggestion
- `delayForCellEdit`: When `true`, Excel queues the batch until the user exits cell edit mode. When `false`, the batch fails immediately if the user is editing. The default value is `false`.
```

docs/excel/excel-add-ins-delay-in-cell-edit.md
```diff
@@ -1,17 +1,42 @@
...
+
+## Behavior comparison
+
+| User editing? | delayForCellEdit = false (default) | delayForCellEdit = true |
```
AlexJerabek:
```suggestion
| User editing? | `delayForCellEdit = false` (default) | `delayForCellEdit = true` |
```

docs/excel/excel-add-ins-delay-in-cell-edit.md
```diff
@@ -1,17 +1,42 @@
...
+  range.values = [["Updated while user was editing elsewhere"]];
+  await context.sync();
+});
 ```
+
+## Guidance
+
+- Use `delayForCellEdit` only when user-initiated commands may overlap active cell editing, such as with a ribbon button that triggers a bulk update.
+- Consider showing a status indicator in your add-in if queued work may be lengthy.
+- Avoid chaining multiple long-running delayed batches. Queue depth can create a perceived lag for your add-in users.
```
AlexJerabek:
```suggestion
- Avoid chaining multiple long-running delayed batches. This creates a perceived lag for your users.
```

docs/excel/excel-add-ins-multiple-ranges.md
```diff
@@ -1,30 +1,35 @@
...
 description: Learn how the Excel JavaScript library enables your add-in to perform operations, and set properties, on multiple ranges simultaneously.
-ms.date: 02/16/2022
+ms.date: 09/22/2025
 ms.localizationpriority: medium
 ---
 
 # Work with multiple ranges simultaneously in Excel add-ins
 
-The Excel JavaScript library enables your add-in to perform operations, and set properties, on multiple ranges simultaneously. The ranges do not have to be contiguous. In addition to making your code simpler, this way of setting a property runs much faster than setting the same property individually for each of the ranges.
+You can apply operations or set properties on several ranges at once, even if they're not contiguous. This keeps code shorter and often runs faster than looping each range separately.
```
AlexJerabek:
```suggestion
You can apply operations or set properties on several ranges at once, even if they're not contiguous. This makes code shorter and more efficient when compared to accessing each range separately.
```

docs/excel/excel-add-ins-multiple-ranges.md
```diff
@@ -1,30 +1,35 @@
...
 
-## RangeAreas
+## Key points
+
+- Use `RangeAreas` to read or set the same thing on several separate ranges in one call.
+- A property is `null` unless all member ranges share the same value.
+- Set a property once on the `RangeAreas` object instead of looping, unless each range needs different logic.
+- Avoid large `RangeAreas` objects made of many single cells. Narrow first with `getSpecialCells` or other filters.
+- Use `getSpecialCells` and `RangeAreas` to target only needed cells before formatting or validation.
```
AlexJerabek: These seem like the same point.

docs/excel/excel-add-ins-multiple-ranges.md
```diff
@@ -78,7 +83,7 @@ The `RangeAreas` type has some properties and methods that are not on the `Range
 
 ## Create RangeAreas
 
-You can create `RangeAreas` object in two basic ways:
+You can create a `RangeAreas` object in two main ways:
```
AlexJerabek: This statement is out of date with the Special Cells APIs. Since we advise using those methods in the key points, we should address them here too. That may require a more significant rewrite of this article.

docs/excel/excel-add-ins-multiple-ranges.md
```diff
@@ -112,6 +117,24 @@ This example applies to scenarios in which you can hard code the range addresses
 - The code runs in the context of a known template.
 - The code runs in the context of imported data where the schema of the data is known.
 
+## Combine `RangeAreas` with `getSpecialCells`
```
AlexJerabek, line 120: I'm not sure this sample shows a good pattern. I think it'd be better to show getting a RangeAreas object through Range.getSpecialCells.

docs/excel/excel-add-ins-multiple-ranges.md
```diff
@@ -141,11 +164,11 @@ await Excel.run(async (context) => {
 });
 ```
 
-Things get more complicated when consistency isn't possible. The behavior of `RangeAreas` properties follows these three principles:
+If values differ, keep these simple rules in mind:
```
AlexJerabek:
```suggestion
Since property values can differ, keep these simple rules in mind.
```

docs/excel/excel-add-ins-multiple-ranges.md
```diff
@@ -141,11 +164,11 @@ await Excel.run(async (context) => {
...
 
-- A boolean property of a `RangeAreas` object returns `false` unless the property is true for all the member ranges.
-- Non-boolean properties, with the exception of the `address` property, return `null` unless the corresponding property on all the member ranges has the same value.
-- The `address` property returns a comma-delimited string of the addresses of the member ranges.
+- A Boolean is `true` only if all ranges are true, otherwise it's `false`.
```
AlexJerabek:
```suggestion
- Boolean properties are `true` only if they're true in all ranges, otherwise they're `false`.
```

docs/excel/excel-add-ins-multiple-ranges.md
```diff
@@ -141,11 +164,11 @@ await Excel.run(async (context) => {
...
+- Other properties are `null` unless all ranges share the same value. (`address` is excluded from this rule.)
+- `address` always returns the comma-delimited addresses string.
```
AlexJerabek: Should we flip these to avoid the awkward exception?
```suggestion
- `address` always returns the comma-delimited addresses string.
- Other properties are `null` unless all ranges share the same value.
```

docs/excel/excel-add-ins-ranges-large.md
```diff
@@ -1,28 +1,66 @@
...
 ms.localizationpriority: medium
 ---
 
 # Read or write to a large range using the Excel JavaScript API
 
-This article describes how to handle reading and writing to large ranges with the Excel JavaScript API.
+Use these patterns to read or write large ranges, while avoiding resource limit errors.
 
-## Run separate read or write operations for large ranges
+## Key points
```
AlexJerabek: I'm not sure you even need a header here.

docs/excel/excel-add-ins-ranges-large.md
```diff
@@ -1,28 +1,66 @@
...
 
-If a range contains a large number of cells, values, number formats, or formulas, it may not be possible to run API operations on that range. The API will always make a best attempt to run the requested operation on a range (i.e., to retrieve or write the specified data), but attempting to perform read or write operations for a large range may result in an API error due to excessive resource utilization. To avoid such errors, we recommend that you run separate read or write operations for smaller subsets of a large range, instead of attempting to run a single read or write operation on a large range.
+- Don't load or write everything at once. Split big ranges into smaller blocks.
```
AlexJerabek:
```suggestion
- Split big ranges into smaller blocks. Don't load or write everything at once.
```

docs/excel/excel-add-ins-ranges-large.md
```diff
@@ -1,28 +1,66 @@
...
+- Load only what you need (for example, just `values` instead of `values,numberFormat,formulas`).
+- Choose row blocks or column blocks based on data shape.
```
AlexJerabek: I'm not sure what this means.

docs/excel/excel-add-ins-ranges-large.md
```diff
@@ -1,28 +1,66 @@
...
+- Use `RangeAreas` to work with scattered cells instead of looping each range.
```
AlexJerabek: Somewhere, this "loop on each range" phrase was introduced, but I'm not sure that's the actual advantage. RangeAreas is also iterating on a set of Ranges, but the advantage is that they are smaller than the whole Range (both in terms of overall cell count and the individual cells per operation).
```suggestion
- Use `getSpecialCells` and `RangeAreas` to work with scattered cells instead of a large range.
```

docs/excel/excel-add-ins-ranges-large.md
```diff
@@ -1,28 +1,66 @@
...
+- If you hit a limit error, retry with a smaller block size.
```
AlexJerabek: Is there a better word than "hit"?

docs/excel/excel-add-ins-ranges-large.md
```diff
@@ -1,28 +1,66 @@
...
-For details on the system limitations, see the "Excel add-ins" section of [Resource limits and performance optimization for Office Add-ins](../concepts/resource-limits-and-performance-optimization.md#excel-add-ins).
+## When to split a large range
 
-### Conditional formatting of ranges
+| Scenario | Sign you should split the range | Approach |
+|----------|----------------------|----------|
+| Reading millions of cells | Timeout or resource error | Read in row or column blocks. Start with 5k–20k rows. |
+| Writing a large result set | Single `values` write fails | Write in row blocks (with same column count for each block). |
+| Sparse updates | Many distant cells | Build combined address string with `getRanges` and `RangeAreas`. |
+| Growing table | Don't know final size | Use `getUsedRange()` then step through new rows until empty. |
```
AlexJerabek: I'm not sure what "growing table" means here. Seems like the add-in would be the thing growing the table, in which case, it would know the size (at least where the end of the last batch was).

docs/excel/excel-add-ins-ranges-large.md
```diff
@@ -1,28 +1,66 @@
...
+| Mixed data and formatting | Formatting slows writes | Write values first, format afterward. |
```
AlexJerabek:
```suggestion
| Writing data and formatting | Formatting slows Excel | Write values first, format afterward. |
```

docs/excel/excel-add-ins-ranges-large.md
```diff
@@ -1,28 +1,66 @@
...
+
+Formatting and calculation-heavy operations, such as conditional formats or formula writes, add time on large areas. Consider:
+
+- First write raw values (plain numbers or text), then add formulas or formats in a second pass.
+- Use `setDirty()` only on necessary recalculation scopes.
+- Limit conditional formats to used rows instead of entire column references (such as `A2:A5000` instead of `A:A`).
+
+## Handling conditional formatting on large data
+
+Avoid applying conditional formatting to entire columns if only part of the column has data. Find the used range and apply the rule there:
```
AlexJerabek: This code sample relates to the last bullet point, so I'm not sure the header and lead-in sentence are needed.

docs/excel/excel-add-ins-ranges-large.md
```diff
@@ -1,28 +1,66 @@
...
+
+```js
+await Excel.run(async (context) => {
+  const sheet = context.workbook.worksheets.getActiveWorksheet();
+  const usedRange = sheet.getUsedRange();
+  usedRange.load("address");
+  await context.sync();
+  const formattedRange = sheet.getRange(usedRange.address.split("!")[1]);
```
AlexJerabek: getUsedRange doesn't return a whole column, so this whole process is unnecessary.

docs/excel/excel-add-ins-ranges-large.md
```diff
@@ -1,28 +1,66 @@
...
+- Compare with patterns for [unbounded ranges](excel-add-ins-ranges-unbounded.md).
+- Explore special cell targeting in [find special cells](excel-add-ins-ranges-special-cells.md).
 
 ## See also
 
 - [Excel JavaScript object model in Office Add-ins](excel-add-ins-core-concepts.md)
 - [Work with cells using the Excel JavaScript API](excel-add-ins-cells.md)
-- [Read or write to an unbounded range using the Excel JavaScript API](excel-add-ins-ranges-unbounded.md)
-- [Work with multiple ranges simultaneously in Excel add-ins](excel-add-ins-multiple-ranges.md)
+- [Apply conditional formatting to Excel ranges](excel-add-ins-conditional-formatting.md)
```
AlexJerabek, line 47: Not sure I see the difference between these two sections. Consider consildating them.

docs/excel/excel-add-ins-ranges-precedents-dependents.md
```diff
@@ -1,18 +1,27 @@
...
 This article provides code samples that retrieve the precedents and dependents of formulas using the Excel JavaScript API. For the complete list of properties and methods that the `Range` object supports, see [Range Object (JavaScript API for Excel)](/javascript/api/excel/excel.range).
 
+## Key points
+
+- Use `getPrecedents` or `getDependents` to get all related cells. Use `getDirectPrecedents` or `getDirectDependents` to get only the first level.
+- Both APIs return a `WorkbookRangeAreas` object. It groups addresses by worksheet. Look in each `areas.items` list for the individual `Range` objects.
+- Very large sets, such as thousands of cells, can be slow. Start with the direct methods or narrow the selection first.
+- These APIs do not cross workbook boundaries.
+- If there are no related cells, an `ItemNotFound` error is thrown. Catch it and show a friendly message.
+- If you highlight cells, give users a way to clear the formatting.
```
AlexJerabek, line 23: Not sure where this tip came from. It's not a bad idea, though undo support has rendered it less important.

docs/excel/excel-add-ins-ranges-precedents-dependents.md
```diff
@@ -119,7 +128,17 @@ await Excel.run(async (context) => {
 ```
 
 > [!NOTE]
-> The `Range.getDependents` and `Range.getDirectDependents` methods return an `ItemNotFound` error if no dependent cells are found.
+> The `Range.getDependents` and `Range.getDirectDependents` methods return an `ItemNotFound` error if no dependent cells are found. Catch this and provide a user-friendly message.
+
+## Best practices for working with precedents and dependents
```
AlexJerabek: This section is redundant. I'd recommend removing the table, since it basically repeats the key points.

docs/excel/excel-add-ins-ranges-special-cells.md
```diff
@@ -1,13 +1,23 @@
...
+
+## Quick reference
+
+| Goal | Use this method | If target might not exist | Result type | Error behavior |
+|------|-----------------|---------------------------|-------------|----------------|
+| Require at least one matching cell | `getSpecialCells` | N/A | `RangeAreas` | Throws `ItemNotFound` if none exist |
+| Optionally act only if matches exist | `getSpecialCellsOrNullObject` | Check `isNullObject` after `context.sync()` | `RangeAreas` proxy | No error, returns `isNullObject = true` |
+
+> [!TIP]
+> Treat `getSpecialCells` like an assertion. Use `getSpecialCellsOrNullObject` when the absence of matches is a valid result, not an error.
```
AlexJerabek, line 20: This is an interesting treatise on *OrNullObject, but I don't think it belongs in this article. Should it be moved to the *OrNullObject article?

docs/excel/excel-add-ins-ranges-special-cells.md
```diff
@@ -21,10 +31,12 @@ getSpecialCells(cellType: Excel.SpecialCellType, cellValueType?: Excel.SpecialCe
 getSpecialCellsOrNullObject(cellType: Excel.SpecialCellType, cellValueType?: Excel.SpecialCellValueType): Excel.RangeAreas;
 ```
 
-The following code sample uses the `getSpecialCells` method to find all the cells with formulas. About this code, note:
+The following code sample uses `getSpecialCells` to find all cells with formulas.
+
+Key points:
```
AlexJerabek: Or could say "Key points:" here too, but having it across multiple lines seems weird to me.
```suggestion
The following code sample uses `getSpecialCells` to find all cells with formulas. Please note:
```

docs/excel/excel-add-ins-ranges-special-cells.md
```diff
@@ -21,10 +31,12 @@ getSpecialCells(cellType: Excel.SpecialCellType, cellValueType?: Excel.SpecialCe
...
 
-- It limits the part of the sheet that needs to be searched by first calling `Worksheet.getUsedRange` and calling `getSpecialCells` for only that range.
-- The `getSpecialCells` method returns a `RangeAreas` object, so all of the cells with formulas will be colored pink even if they are not all contiguous.
+- Restrict the search scope for better performance by calling `worksheet.getUsedRange()` first.
```
AlexJerabek: Rephrasing so this is about the code and not a best practice.
```suggestion
- The search scope is restricted for better performance by calling `worksheet.getUsedRange()` first.
```

docs/excel/excel-add-ins-ranges-special-cells.md
```diff
@@ -115,9 +129,13 @@ await Excel.run(async (context) => {
 });
 ```
 
+## Next steps
+
+- Combine special-cell queries with [string search](excel-add-ins-ranges-string-match.md) for richer auditing.
+- Apply formatting, comments, or data validation to the resulting `RangeAreas`.
+- Chain with [work with multiple ranges](excel-add-ins-multiple-ranges.md) to process disjoint areas efficiently.
```
AlexJerabek: This is essentially the same tip. Also, is "disjoint" the right word?
```suggestion
- Apply formatting, comments, or data validation to the resulting [RangeAreas](excel-add-ins-multiple-ranges.md) to process disjoint areas efficiently.
```

docs/excel/excel-add-ins-ranges-special-cells.md
```diff
@@ -115,9 +129,13 @@ await Excel.run(async (context) => {
 });
 ```
 
+## Next steps
+
+- Combine special-cell queries with [string search](excel-add-ins-ranges-string-match.md) for richer auditing.
```
AlexJerabek, line 132: I'm not sure what this scenario would be. I agree that string search is a useful related concept,

docs/excel/performance.md
```diff
@@ -1,25 +1,37 @@
...
 ---
 
 # Performance optimization using the Excel JavaScript API
 
-There are multiple ways that you can perform common tasks with the Excel JavaScript API. You'll find significant performance differences between various approaches. This article provides guidance and code samples to show you how to perform common tasks efficiently using Excel JavaScript API.
+Write faster, more scalable Excel add-ins by minimizing processes, batching functions, and reducing payload size. This article shows patterns, anti-patterns, and code samples to help you optimize common operations.
+
+## Quick improvements
+
+Apply these strategies first for the largest immediate impact:
```
AlexJerabek:
```suggestion
Apply these strategies first for the largest immediate impact.
```

docs/excel/performance.md
```diff
@@ -62,22 +74,27 @@ await Excel.run(async (context) => {
 });
 ```
 
-Please note that only formula calculations are suspended. Any altered references are still rebuilt. For example, renaming a worksheet still updates any references in formulas to that worksheet.
+Only formula calculations are suspended. Structural updates, such as renaming a worksheet, still occur.
```
AlexJerabek: The meaning changed here. The point is that renaming the workbook updates any dependent formulas, even if calculations are suspended.

docs/excel/performance.md
```diff
@@ -62,22 +74,27 @@ await Excel.run(async (context) => {
...
 
 ## Importing data into tables
 
-When trying to import a huge amount of data directly into a [Table](/javascript/api/excel/excel.table) object directly (for example, by using `TableRowCollection.add()`), you might experience slow performance. If you are trying to add a new table, you should fill in the data first by setting `range.values`, and then call `worksheet.tables.add()` to create a table over the range. If you are trying to write data into an existing table, write the data into a range object via `table.getDataBodyRange()`, and the table will expand automatically.
+When you import large datasets directly into a [Table](/javascript/api/excel/excel.table), such as repeatedly calling `TableRowCollection.add()`, performance can degrade. Instead, take the following approach:
+
+1. Write the entire 2D array to a range with `range.values`.
+2. Create the table over that populated range (`worksheet.tables.add()`).
+
+For existing tables, set values on `table.getDataBodyRange()` in bulk. The table expands automatically.
```
AlexJerabek, line 97: I think this entire tip is a relic from ExcelApi 1.1. You can add multiple rows at once as of 1.4, which doesn't seem less efficient than the tips here.

docs/excel/performance.md
```diff
@@ -107,19 +124,19 @@ await Excel.run(async (context) => {
...
 
-The payload size of a request is a combination of the following three components.
+The payload size of a request combines:
 
-* The number of API calls
-* The number of objects, such as `Range` objects
-* The length of the value to set or get
+- The number of API calls
+- The number of objects, such as `Range` objects
+- The length of the value to set or get
```
AlexJerabek: Do these bulleted list items need periods after them?

docs/excel/performance.md
```diff
@@ -239,9 +256,14 @@ async function run() {
 }
 ```
 
+## Next steps
+
+- Review [Resource limits and performance optimization](../concepts/resource-limits-and-performance-optimization.md) for host-level constraints.
+- Explore [Working with multiple ranges](excel-add-ins-multiple-ranges.md) to create fewer objects.
+- Add telemetry (operation durations, row counts) to guide further performance optimization.
```
AlexJerabek:
```suggestion
- Add telemetry for data such as operation durations and row counts to guide further performance optimization.
```

AlexJerabek (changes requested):

docs/excel/excel-add-ins-ranges-large.md
alison-mk: Do you mean throughout articles, or just in this instance? 

My take is that they're different sections. All our articles have "See also" so that's a consistent format for readers, but "Next steps" is a new format. "Next steps" gives sentence-based tips, while "See also" is just a list of articles. I did make sure links aren't repeated between the two setions.

docs/excel/excel-add-ins-ranges-precedents-dependents.md
alison-mk: I think it's helpful, so I'm inclined to keep it unless you think otherwise.

docs/excel/excel-add-ins-data-validation.md
alison-mk: These changes (here and in the related PRs) add "Key points" in multiple places. I agree that the "Please note" is more consistent with existing tone, but I think "Key points" is more direct. I'm inclined to keep "Key points" unless you feel strongly!

docs/excel/excel-add-ins-data-validation.md
alison-mk: I can think of theoretical scenarios -- something like this? An event happens in the workbook (perhaps `onFormulaChanged` or `onColumnSorted`), then an add-in uses validation to check the formula or formula results in affected cells

docs/excel/excel-add-ins-ranges-large.md
alison-mk: Agreed. Jargon accidentally made it through! Will remove.

docs/excel/excel-add-ins-ranges-large.md
alison-mk: I see your point. Will remove this line.

docs/excel/excel-add-ins-ranges-large.md
alison-mk: Clarifying because this combined line comment formatting is new -- Does your comment mean we should get rid of the entire sample here? Lines 47-50 are the entire sample, minus line 46 (which has getActiveWorksheet)

docs/excel/excel-add-ins-ranges-special-cells.md
alison-mk: Do we need to have a specific scenario in mind? Useful related concept seems compelling enough to me

docs/excel/excel-add-ins-ranges-special-cells.md
alison-mk: Hot take: This new multi-line comment feature is confusing. 

Related to your comment: I see your point. I'll move it.

docs/excel/excel-add-ins-ranges-precedents-dependents.md
alison-mk: I agree that some lines are redundant and can remove it -- but an opposing perspective (why I included the table in this PR) is that this table targets a different learning style than the Key points list.

alison-mk (commented):
Thanks for reviewing so quickly and providing helpful feedback! A few questions in response.

docs/excel/excel-add-ins-data-validation.md
AlexJerabek: That's fair. We can keep it!

docs/excel/excel-add-ins-data-validation.md
AlexJerabek: That makes sense. Offhand, I'm not sure if DataValidation and formulas work together, but I guess I have no reason to suspect otherwise.

docs/excel/excel-add-ins-ranges-large.md
AlexJerabek: The issue is that I don't think this will generate a meaningful RangeAreas. It scopes it to the used range, but that's still just a single range.

docs/excel/excel-add-ins-ranges-large.md
AlexJerabek: As a general statement, but I get your point. We can leave as is.

docs/excel/excel-add-ins-ranges-precedents-dependents.md
AlexJerabek: Fair. I guess I'm on the fence because it doesn't seem like a particularly long article, so having a Key Points summary and the table feels like overkill. Your call.

docs/excel/excel-add-ins-ranges-special-cells.md
AlexJerabek: Yeah... it's good for suggestions, but bad otherwise. Noted!

docs/excel/excel-add-ins-ranges-special-cells.md
AlexJerabek: No, but I want to make sure we don't lead people down unsupported or frustrating paths. Like, if string search is internally optimized, are we encouraging adding an unnecessary filtering step beforehand?

docs/excel/excel-add-ins-ranges-large.md
alison-mk: I see your point. I had tested the code sample to make sure it worked (no error, does what it says) but I didn't catch the major flaw that you identified. I removed this sample and added a note to use `getUsedRange` in the preceding section.

docs/excel/excel-add-ins-ranges-precedents-dependents.md
alison-mk: I agree. I'll remove this table, but I do think targeting various learning styles with is something for us to keep in mind for future article revisions.

docs/excel/excel-add-ins-data-validation.md
alison-mk: I tested this with the onRowSorted event and Range.dataValidation.clear() -- it works. I'll leave this line in.

docs/excel/excel-add-ins-multiple-ranges.md
alison-mk: This article has an existing section about using getSpecialCells (Get special cells from multiple ranges) and lists the getSpecialCells APIs in the methods list earlier in the article. It seems like there are many ways to create RangeAreas -- as illustrated by the methods list -- and this section is simply describing how to create RangeAreas as a broader concept. Perhaps I'm misunderstanding? If so, this is a longer conversation that we could have offline. I suspect that it doesn't block merging this PR, let me know if otherwise.

docs/excel/excel-add-ins-multiple-ranges.md
alison-mk: Could you elaborate on why? Is getting a RangeAreas object and then using getSpecialCells on that RangeAreas an inefficient strategy? 

To clarify, are you suggesting a sample that does the opposite -- uses getSpecialCells on a Range, and then returns a RangeAreas object of the special cells? 

If yes to both of those questions, then I think this section should just be removed and the following section with a link to the special cells article is sufficient.

docs/excel/performance.md
alison-mk: Do you think the entire "Importing data into tables" section can be removed from this Performance article?

docs/excel/excel-add-ins-ranges-special-cells.md
alison-mk: Ok, I've come up with a scenario. String search returns the first cell with matching text (including some search parameters). An add-in developer may want to first narrow the search parameters in a way that's not supported by string [SearchCriteria](https://learn.microsoft.com/en-us/javascript/api/excel/excel.searchcriteria?view=excel-js-preview) -- such as only searching cells with the same data validation or same conditional format. I will leave this line, unless you disagree!

docs/excel/excel-add-ins-ranges-special-cells.md
alison-mk: I have an update for this thread. The [existing article](https://learn.microsoft.com/en-us/office/dev/add-ins/excel/excel-add-ins-ranges-special-cells) has a detailed description of the *OrNullObject impact on getSpecialCells. This new article section provides an introduction to that existing content. Does it seem more relevant to you when interpreted that way?

alison-mk (commented):
Follow up questions or responses for all of the remaining open discussions. Thanks for the great feedback!

docs/excel/excel-add-ins-multiple-ranges.md
AlexJerabek: My comment is more specific. The next two lines are Worksheet.getRanges and Worksheet.getSelectedRanges, but there are more ways to create RangeAreas (and arguably better ones). I think these were the two ways before more functions were added in a later API set. 

I think this section could either be rewritten to include all the ways to generate a RangeAreas (or at least add getSpecialCells and dependants/precedents), or be changed to a section about using Worksheet.getRanges and add another section about getSelectedRanges (that way, it's organized by "getting RangeAreas through string addresses", "getting RangeAreas through user selection", and "getting RangeAreas through cell criteria").

docs/excel/excel-add-ins-multiple-ranges.md
AlexJerabek: Thanks for calling me out on this. You're right that there's nothing wrong with this sample. Originally, I thought you should just show calling getSpecialCells on A1:C100, but there are reasons to avoid the B column. Please ignore this.

docs/excel/excel-add-ins-ranges-special-cells.md
AlexJerabek: That's fair. Thank you.

docs/excel/excel-add-ins-ranges-special-cells.md
AlexJerabek: Sounds good to me!

docs/excel/performance.md
AlexJerabek: Yeah, that's my understanding. Might be worth asking someone like Joe, but don't let that block your check-in.

docs/excel/excel-add-ins-multiple-ranges.md
alison-mk: Thanks for the details! I added Worksheet.getSpecialCells to this list. 

I considered adding Range.getPrecedents (and the other similar methods) like you suggest, but I noticed that they return a WorkbookRangeAreas object--not a RangeAreas object. WorkbookRangeAreas are not currently addressed anywhere in this article.

That seems like an oversight. I think WorkbookRangeAreas should be addressed in this article, but I'd call that rewrite outside the scope of this low engagement review PR. I'll make a task to do some research internally and then do a full audit of this multiple ranges article. Since you've approved, I'll merge this PR within the next 24 hours. Thanks again for all the great feedback!

PR #5457, opened 2025-11-05
[All hosts] (event-based activation) Update Cache-Control header guidance

samantharamon:

PR #5465, opened 2025-11-07
Revert "[All hosts] (event-based activation) Update Cache-Control header guidance"

samantharamon:
After receiving customer feedback and doing further testing, this PR reverts the changes in OfficeDev/office-js-docs-pr#5457. The guidance on not using `Cache-Control` headers during development still applies.

### DIFF
--- a/docs/outlook/add-and-remove-attachments-to-an-item-in-a-compose-form.md
+++ b/docs/outlook/add-and-remove-attachments-to-an-item-in-a-compose-form.md
@@ -1,7 +1,7 @@
 ---
 title: Add and remove attachments in an Outlook add-in
 description: Use various attachment APIs to manage the files or Outlook items attached to the item the user is composing.
-ms.date: 11/07/2025
+ms.date: 03/12/2025
 ms.topic: how-to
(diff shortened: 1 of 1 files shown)


docs/outlook/add-and-remove-attachments-to-an-item-in-a-compose-form.md
```diff
@@ -48,7 +48,7 @@ You can attach a file to a message or appointment in a compose form by using the
 
 > [!NOTE]
 >
-> - The URI of the file to be attached must support caching in production. The server hosting the image shouldn't return a `Cache-Control` header that specifies `no-cache`, `no-store`, or similar options in the HTTP response. However, when you're developing the add-in and making changes to files, caching can prevent you from seeing your changes. We recommend using `Cache-Control` headers during development.
```
timwan10, line 51: We should leave this in, as I didn't test/verify this. Let's just not touch this.

samantharamon: Reverted this change.

PR #5535, opened 2025-12-05
[Excel] (Custom functions) Add synchronous support article

alison-mk:
This PR documents how to support synchronous Excel processes in custom functions add-ins.

### DIFF
--- a/docs/excel/custom-functions-synchronous.md
+++ b/docs/excel/custom-functions-synchronous.md
@@ -7,16 +7,16 @@ ms.localizationpriority: medium
 
 # Synchronous custom functions
 
-Synchronous custom functions allow evaluate and conditional format processes to run in Excel simultaneously with the custom function. Enable synchronous support when your custom function needs to run in tandem with any of the Excel processes listed in this article. Custom functions that don't support synchronous operations return an error such as `#CALC!` or `#VALUE!` when they run at the same time as these Excel processes.
+Synchronous custom functions enable the evaluation and conditional format processes to run in Excel simultaneously with the custom function. Enable synchronous support when your custom function needs to run in tandem with any of the Excel processes listed in this article. Custom functions that don't support synchronous operations return an error such as `#CALC!` or `#VALUE!` when they run at the same time as these Excel processes.
 
 > [!WARNING]
-> Synchronous custom functions don't support write operations with Office JavaScript APIs, such as using `Range.values` to set a cell value. Calling a write operation in a synchronous custom function may cause Excel to freeze. Do not use write operations in synchronous custom functions.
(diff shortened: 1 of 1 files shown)


docs/excel/custom-functions-synchronous.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Synchronous custom functions
+description: Use synchronous custom functions to support evaluation and conditional format processes in Excel.
+ms.date: 12/09/2025
+ms.localizationpriority: medium
+---
+
+# Synchronous custom functions
+
+Synchronous custom functions allow evaluate and conditional format processes to run in Excel simultaneously with the custom function. Enable synchronous support when your custom function needs to run in tandem with any of the Excel processes listed in this article. Custom functions that don't support synchronous operations return an error such as `#CALC!` or `#VALUE!` when they run at the same time as these Excel processes.
```
Rick-Kirkham: Is it standard in the custom functions world to use the verb "evaluate" as a noun? Can we use the noun instead? As per style guide, use "enable" not "allow".

```suggestion
Synchronous custom functions enable the evaluation and conditional format processes to run in Excel simultaneously with the custom function. Enable synchronous support when your custom function needs to run in tandem with any of the Excel processes listed in this article. Custom functions that don't support synchronous operations return an error such as `#CALC!` or `#VALUE!` when they run at the same time as these Excel processes.
```

docs/excel/custom-functions-synchronous.md
```diff
@@ -0,0 +1,82 @@
...
+
+> [!WARNING]
+> Synchronous custom functions don't support write operations with Office JavaScript APIs, such as using `Range.values` to set a cell value. Calling a write operation in a synchronous custom function may cause Excel to freeze. Do not use write operations in synchronous custom functions.
```
Rick-Kirkham: Last sentence is redundant.
```suggestion
> Synchronous custom functions don't support write operations with Office JavaScript APIs, such as using `Range.values` to set a cell value. Calling a write operation in a synchronous custom function may cause Excel to freeze.
```

docs/excel/custom-functions-synchronous.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Excel processes supported by synchronous custom functions
+
+The following actions and processes work with synchronous custom functions.
+
+### Supported evaluate actions
```
Rick-Kirkham: better?
```suggestion
### Supported evaluation actions
```

docs/excel/custom-functions-synchronous.md
alison-mk: Great feedback, thank you!

PR #5555, opened 2025-12-18
[Excel] (Custom functions) Update custom enum schema

alison-mk:
The custom enum schema was updated to divide one property (`value`) into two properties (`numberValue` and `stringValue`). This schema change is backwards compatible -- functions using the old schema will still work as expected when it was developed.

### DIFF
--- a/docs/excel/custom-functions-json.md
+++ b/docs/excel/custom-functions-json.md
@@ -189,8 +189,8 @@ The `enums` property is an array of [enum](https://www.typescriptlang.org/docs/h
 | :------------ | :-------- | :------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `name` | string | Yes | A brief description of the constant. |
 | `tooltip` | string | No | Additional information about the constant that can be shown as a tooltip in user interfaces. |
-| `numberValue` | number | Yes | The number value of the constant. If the enum `type` is `number`, then this field is required. |
-| `stringValue` | string | Yes | The string value of the constant. If the enum `type` is `string`, then this field is required. |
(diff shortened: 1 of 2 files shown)


docs/excel/custom-functions-json.md
```diff
@@ -187,9 +187,10 @@ The `enums` property is an array of [enum](https://www.typescriptlang.org/docs/h
 
 | Property      | Data type | Required | Description                                                                                                                                                                      |
 | :------------ | :-------- | :------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `name` | string    | Yes       | A brief description of the constant. |
-| `tooltip` | string    | No       | Additional information about the constant that can be shown as a tooltip in user interfaces. |
-| `value` | string    | Yes      | The value of the constant. |
+| `name` | string | Yes | A brief description of the constant. |
+| `tooltip` | string | No | Additional information about the constant that can be shown as a tooltip in user interfaces. |
+| `numberValue` | number | Conditional | The number value of the constant. If the enum `type` is `number`, then this field is required. |
+| `stringValue` | string | Conditional | The string value of the constant. If the enum `type` is `string`, then this field is required. |
```
AlexJerabek, line 193: What happens if the enum `type` is neither number nor string? Is that possible?

AlexJerabek (approved):
Approving with an open question

docs/excel/custom-functions-json.md
alison-mk: Good question! I will find out and clarify this in another PR. For now, I'll merge this PR.

PR #5584, opened 2026-01-22
[excel] (Custom Functions) Remove over-communication of version support

AlexJerabek:


docs/excel/custom-functions-authentication.md
```diff
@@ -9,8 +9,6 @@ ms.localizationpriority: medium
 
 In some scenarios, a custom function that doesn't use a [shared runtime](../testing/runtimes.md#shared-runtime) will need to authenticate the user in order to access protected resources. Custom functions that don't use a shared runtime run in a [JavaScript-only runtime](../testing/runtimes.md#javascript-only-runtime). Because of this, if the add-in has a task pane, you'll need to pass data back and forth between the JavaScript-only runtime and the HTML-supporting runtime used by the task pane. You do this by using the [OfficeRuntime.storage](/javascript/api/office-runtime/officeruntime.storage) object and a special Dialog API.
 
-[!include[Excel custom functions note](../includes/excel-custom-functions-note.md)]
```
alison-mk, line 12: This include file contains two notes. Do you think we should remove the unified manifest note, too? 

image

AlexJerabek: My thought was that the manifest information makes sense in the overview and troubleshooting articles but is overkill in every article. It also might make sense in [this requirement set article](https://learn.microsoft.com/en-us/javascript/api/requirement-sets/excel/custom-functions-requirement-sets).

alison-mk: Thanks for explaining! I see now how you approached it.

PR #5595, opened 2026-02-06
[All Hosts] (troubleshooting) Errors in 'ADD-IN INFORMATION' box.

Rick-Kirkham:
Fulfills a customer request.


docs/images/add-in-information-box.png
alison-mk, line 1: The text in this image is blurry, even when zooming in, and would be difficult to read for low vision users. To meet WCAG accessibility guidelines, a user should be able to zoom up to 200% to enlarge text. With a screenshot of text, this means the image should be high enough resolution to allow larger and clearer text with zooming in. Basically, the resolution of this image is too low. 

I realize this screenshot might be hard to acquire because the scenario is so niche, so I did a little tweaking and sent you a version with higher resolution offline.

alison-mk: Here's a higher resolution image that allows zooming to enlarge the text: image

Issue #5187, opened 2025-05-20
Feedback - Unclear of needed steps for typical initial SSO configuration for Office.js Addin


thekingofspain:
### Type of issue

Other (describe below)

### Feedback

I am trying to setup up an Office.js addin that has permission to getAccessToken.  I am looking for the required step needed to call this api.  This page is very confusing and does not follow the typical enterprise steps.

1.  Most developers in an enterprise do not have access to the O365 admin center or Entra.
2.  The initial Office.js Addin Integrated app registration creates the initial Entra application registration from the manifest file that the developer provides..
3.  The person registering the initial Office.js addin, needs to go to Entra and get the create the Application (client) Id for the developer.
4.  The developer needs to change the addin manifest.xml.
- Increment the version
- Add the WebApplicationInfo element in:
-- Id
-- Resource
-- Scopes - This should only mention and list the needed scopes to call getAccessToken, nothing else.
5.  The O365 or Entra Admin need to update the Entra configuration to allow the Office.js addin to call the getAccessToken api.
6.  After the Entra app registration is updated, the O365 Admin must update the addin registration in O365 admin center with the newly update manifest from the developer.

 Issues:
- There needs to a section/tab/page for updating an existing Entra application registration to support getAccessToken.
- Need Scopes
   -- Manifest.xml scopes
   -- Entra Scopes
       OpenId and Profile mentions have different counts and OpenId is not listed as an added scope in the existing Entra configuration step.
   --  Remove Extra scopes (Outlook, User.Read, etc., that are not required for getAccessToken)
  - Entra Screen shots per needed menu page with values to check highlighted.
  - Entra Manifest values, what values need to be set in the Entra manifest for the addin to call getAccessToken.
  - Entra Screen Shots to Manifest values.
  - Are SSO OAuth urls needed or not?
  - Should a secret be added?

Cross referencing this page to other SSO based office.js pages provide similar but different configuration steps.  The getAccessToken page should be the definitive page on how to setup Entra from new Entra application registration in addition to updating an Entra App registration created via the O365 admin center.

### Page URL

https://learn.microsoft.com/en-us/office/dev/add-ins/develop/use-sso-to-get-office-signed-in-user-token?tabs=vs2019

### Content source URL

https://github.com/OfficeDev/office-js-docs-pr/blob/main/docs/develop/use-sso-to-get-office-signed-in-user-token.md

### Author

@o365devx

### Document Id

[...]

PR #5644, opened 2026-03-09
[All hosts](SSO) Update app registration steps

davidchesnut:
- Removes VS steps. Only yo office shown.
- Updates for changes in the portal UI.


davidchesnut:
fixes #5187

PR #5663, opened 2026-03-16
[All Hosts] (engagement) Refresh 8 devx articles

Rick-Kirkham:


docs/develop/support-for-task-pane-and-content-add-ins.md
```diff
@@ -108,19 +108,19 @@ For detailed information about how to manage custom XML parts with a task pane a
...
+To avoid roundtrips to the server where the document is stored, the add-in manages data created by using the `Settings` object in memory at run time. When the add-in initializes, it loads previously saved settings data into memory. The add-in saves changes to that data back to the document only when you call the [Settings.saveAsync](/javascript/api/office/office.settings#office-office-settings-saveasync-member(1)) method. Internally, the data is stored in a serialized JSON object as name/value pairs. Use the [get](/javascript/api/office/office.settings#office-office-settings-get-member(1)), [set](/javascript/api/office/office.settings#office-office-settings-set-member(1)), and [remove](/javascript/api/office/office.settings#office-office-settings-remove-member(1)) methods of the **Settings** object to read, write, and delete items from the in-memory copy of the data. The following line of code shows how to create a setting named `themeColor` and set its value to 'green'.
```
alison-mk: Per term studio, "round trip" should be two words 
```suggestion
To avoid round trips to the server where the document is stored, the add-in manages data created by using the `Settings` object in memory at run time. When the add-in initializes, it loads previously saved settings data into memory. The add-in saves changes to that data back to the document only when you call the [Settings.saveAsync](/javascript/api/office/office.settings#office-office-settings-saveasync-member(1)) method. Internally, the data is stored in a serialized JSON object as name/value pairs. Use the [get](/javascript/api/office/office.settings#office-office-settings-get-member(1)), [set](/javascript/api/office/office.settings#office-office-settings-set-member(1)), and [remove](/javascript/api/office/office.settings#office-office-settings-remove-member(1)) methods of the **Settings** object to read, write, and delete items from the in-memory copy of the data. The following line of code shows how to create a setting named `themeColor` and set its value to 'green'.
```

PR #5680, opened 2026-03-23
Lindalu update known issues

lindalu-MSFT:


docs/resources/resources-office-add-in-known-issues.md
```diff
@@ -1,68 +1,82 @@
 ---
 title: Office Add-ins known issues
 description: This article documents active and resolved issues with Office Add-ins.
-ms.date: 03/18/2026
+ms.date: 03/23/2026
 ms.localizationpriority: medium
 ---
 
-# Office Add-ins known issues
+# Active issues in Office add-ins
```
AlexJerabek, line 8: Why did we rename the page? Is this term coming up more in internal communcations?

AlexJerabek: Also, the TOC still says "Known issues". There might be other links with the old title too.

image

lindalu-MSFT: > Why did we rename the page? Is this term coming up more in internal communcations?

@AlexJerabek I changed the H1 because there was a nearly identical H2 immediately following. This pushed the headings down a level unnecessarily. The actual issue title did not appear in the topic TOC in the right pane and readers couldn't easily scan the issues. By deleting one of the headings and elevating each heading after, the Issues became H2 instead of H3 so that was fixed. I changed it to the original H1 instead of the H2.

image

WAS
# Office Add-ins Known issues
## Active issues in Office add-ins

I CHANGED TO
# Active issues in Office add-ins (deleted the H1)

REVERTED BACK TO
# Office Add-ins Known issues (original H1)

You're right, I reverted to the original H1 so it does sync with the file name and the title of the page as you suggest. Also, the page contains Active AND Resolved issues so the H1 is the better choice.

lindalu-MSFT: Task: https://office.visualstudio.com/OC/_workitems/edit/11341758

PR #5698, opened 2026-03-30
[all hosts] (Dialog) Explicitly state window.open should not be called

AlexJerabek:
We imply in the docs to only use our dialog, rather than having the task pane open new windows, but we don't explicitly say that. This PR adds that stipulation. It also calls out some patterns around users blocking pop-ups.

### DIFF
--- a/docs/develop/dialog-best-practices.md
+++ b/docs/develop/dialog-best-practices.md
@@ -77,7 +77,7 @@ Your add-in should always handle error 12009 gracefully. Here are some recommend
         } else {
           // Dialog opened successfully.
           const dialog = asyncResult.value;
-          // ... handle dialog events
+          // ... Handle dialog events here.
         }
       }
     );
@@ -85,21 +85,7 @@ Your add-in should always handle error 12009 gracefully. Here are some recommend
 
 - Provide an alternative workflow when possible. If your add-in can function with reduced capabilities when the dialog is blocked, offer that alternative to the user.
 
-    ```javascript
-    if (asyncResult.error.code === 12009) {
-      const userChoice = confirm(
-        "This feature requires a dialog to sign you in. " +
-        "Would you like to try again? Click 'Cancel' to continue with limited features."
-      );
-      if (userChoice) {
-        // User wants to retry - show instructions.
-        showRetryInstructions();
-      } else {
-        // Provide limited functionality without authentication.
-        continueWithoutAuth();
-      }
-    }
-    ```
+    
 
 #### Disabling the prompt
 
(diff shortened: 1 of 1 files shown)


docs/develop/dialog-best-practices.md
```diff
@@ -145,23 +200,23 @@ function openFirstDialog() {
...
 
-So, if you passed a route to the `displayDialogAsync` method, you wouldn't really have an SPA; you'd have *two instances of the same SPA*. Moreover, much of the code in the task pane instance would never be used in that instance and much of the code in the dialog box instance would never be used in that instance. It would be like having two SPAs in the same bundle.
+So, if you pass a route to the `displayDialogAsync` method, you don't really have an SPA; you have *two instances of the same SPA*. Moreover, much of the code in the task pane instance is never used in that instance and much of the code in the dialog box instance is never used in that instance. It's like having two SPAs in the same bundle.
 
 #### Microsoft recommendations
 
-Instead of passing a client-side route to the `displayDialogAsync` method, we recommend that you do one of the following:
+Instead of passing a client-side route to the `displayDialogAsync` method, use one of the following approaches:
```
samantharamon:
```suggestion
Instead of passing a client-side route to the `displayDialogAsync` method, use one of the following approaches.
```

docs/develop/dialog-best-practices.md
```diff
@@ -45,38 +45,93 @@ For best practices in dialog box design, see [Dialog boxes in Office Add-ins](..
...
+              "Please click the add-in button again and choose 'Allow' when prompted."
+            );
+          } else {
+            // Handle other errors.
+            showNotification("Error", `Unable to open dialog: ${asyncResult.error.message}`);
+          }
+        } else {
+          // Dialog opened successfully.
+          const dialog = asyncResult.value;
+          // ... handle dialog events
```
samantharamon:
```suggestion
          // ... Handle dialog events here.
```

docs/develop/dialog-best-practices.md
```diff
@@ -45,38 +45,93 @@ For best practices in dialog box design, see [Dialog boxes in Office Add-ins](..
...
+      );
+      if (userChoice) {
+        // User wants to retry - show instructions.
+        showRetryInstructions();
+      } else {
+        // Provide limited functionality without authentication.
+        continueWithoutAuth();
+      }
+    }
+    ```
```
samantharamon:
```suggestion
- Provide an alternative workflow when possible. If your add-in can function with reduced capabilities when the dialog is blocked, offer that alternative to the user.

    ```javascript
    if (asyncResult.error.code === 12009) {
      // Provide limited functionality without authentication.
      continueWithoutAuth();
    }
    ```
```
I'm not sure if `window.confirm` reliably works in an Office Add-in. It would also likely be blocked by pop-up blockers, so it may not be the best method to provide an alternative workflow.

AlexJerabek: Good call. Much simpler code snippet too.

PR #5735, opened 2026-04-21
[Excel] (custom functions) clarify lack of support in manifest conversion

Rick-Kirkham:


alison-mk (approved):
Thanks for adding this!

PR #5757, opened 2026-05-07
[excel] (Tables) Remove requirement set check in table sample

AlexJerabek:
This PR also cleans up some style issues.


davidchesnut (approved):
Nice!

PR #5764, opened 2026-05-13
[All hosts] (dialog) Refresh article

samantharamon:


docs/develop/parent-to-dialog.md
```diff
@@ -1,24 +1,33 @@
...
-  [!INCLUDE [browser-security-updates](../includes/browser-security-updates.md)]
+For the recommended approach using `messageChild`, see [Pass information to the dialog box](dialog-api-in-office-add-ins.md#pass-information-to-the-dialog-box).
 
-- Add query parameters to the URL that is passed to `displayDialogAsync`.
+## Choose an approach
+
+| Consideration | Local storage | Query parameters |
+| --- | --- | --- |
+| **Data size** | Suitable for larger payloads | Best for small values |
+| **Timing** | Available anytime after the dialog opens | Available only on initial load |
```
Rick-Kirkham: "Available only on initial load" seems misleading to me. The values could be read and assigned to variables which can themselves be read as long as the dialog runtime remains open.

docs/develop/parent-to-dialog.md
```diff
@@ -1,24 +1,33 @@
...
+| **Persistence across navigation** | Persists until explicitly cleared | Lost if the dialog redirects |
+| **Same-origin requirement** | Host page and dialog must share the same origin | Host page and dialog must share the same origin |
```
Rick-Kirkham: Since the value of both columns is the same, does this row serve any purpose?

Rick-Kirkham (approved):
Approving with a couple of comments.

docs/develop/parent-to-dialog.md
samantharamon: Updated.

docs/develop/parent-to-dialog.md
samantharamon: Removed.

PR #5778, opened 2026-06-03
[Excel] (Custom functions) Additional unified manifest notes

alison-mk:
Adding another preview tag and a minor clarification of XML-only support.


davidchesnut (approved):
Looks good!

PR #5788, opened 2026-06-11
[PowerPoint] (Tables) Add helper function to cell property creation snippet

AlexJerabek:
There's a tricky piece of code to create a 2D array full of unique objects. This PR pulls that into a helper function that's referenced elsewhere in the sample.


davidchesnut (approved):
Looks good!

PR #5798, opened 2026-06-17
[excel] (Custom Functions) Reorganize CF Auth article for better engagement

AlexJerabek:
This PR reorganizes the CF Auth article for engagement. It pulls the workflow bulletpoints to the beginning and groups the Dialog and Storage samples more cohesively.


davidchesnut (approved):
Looks great!

PR #5804, opened 2026-06-24
[Excel] (copilot) create Excel Copilot skills

Rick-Kirkham:
This is a new way to use Office.js: in a custom Copilot skill, not an add-in. 
The article mainly relies on another article in the Copilot doc set. It adds supplementary information to that article. 
This will be improved soon, but we needed something for an announcement tomorrow.

### DIFF
--- a/docs/excel/excel-skills.md
+++ b/docs/excel/excel-skills.md
@@ -22,7 +22,9 @@ Start with the section [What you'll build](/microsoft-365/copilot/cowork/cowork-
 
 ## Supplement to Step 1
 
-1. In the YAML frontmatter of a SKILL.md file, include a `metadata.tags` property with an `excel` tag, and other tags as needed. The following is an example.
+Follow these steps to supplement [Step 1: Create your first skill](/microsoft-365/copilot/cowork/cowork-plugin-development#build-a-plugin-from-scratch#step-1-create-your-first-skill).
+
+1. In the YAML frontmatter of a `SKILL.md` file, include a `metadata.tags` property with an `excel` tag, and other tags as needed. The following is an example.
 
    ```yaml
    name: my-excel-skill
@@ -33,7 +35,7 @@ Start with the section [What you'll build](/microsoft-365/copilot/cowork/cowork-
      tags: excel
    ```
 
-1. For most Excel skills, the output should be in the workbook, not a chat or CLI interface. So, in place of the "Output Format" section of a SKILL.md file, have a "Workbook Output" section. The following is an example.
+1. For most Excel skills, the output should be in the workbook, not a chat or CLI interface. So, in place of the "Output Format" section of a `SKILL.md` file, have a "Workbook Output" section. The following is an example.
 
    ```md
    ## Workbook output
@@ -44,7 +46,7 @@ Start with the section [What you'll build](/microsoft-365/copilot/cowork/cowork-
    - ... other characteristics here.
    ```
 
-1. Include a "Common pitfalls to avoid" section in the SKILL.md. The following is an example.
+1. Include a "Common pitfalls to avoid" section in the `SKILL.md`. The following is an example.
 
    ```md
    ## Common pitfalls to avoid
@@ -68,7 +70,7 @@ Start with the section [What you'll build](/microsoft-365/copilot/cowork/cowork-
    > [!IMPORTANT]
    > The script should *not* call `Office.onReady` or define an `Office.initialize`. Copilot in Excel automatically creates a runtime and initializes Office.js. 
 
-1. Include as resources, two markdown files that will impose some boundaries on how the skill interacts with the workbook. 
+1. Include two markdown files as resources that will impose some boundaries on how the skill interacts with the workbook. 
 
    1. The first resource should give instructions on how to the existing workbook should be cleaned up or normalized before the skill adds any output to it. These instructions are specific to the scenario the skill addresses, not generic Excel best practices. The following is an example from a skill that creates a dashboard with data about an athletic rivalry between two American colleges.
 
@@ -105,7 +107,7 @@ Start with the section [What you'll build](/microsoft-365/copilot/cowork/cowork-
    
    1. The second resource should do the following.
 
-      - Provide some general quality rules that Copilot should always following when using the skill.
+      - Provide some general quality rules that Copilot should always follow when using the skill.
       - Explain when the skill's scripts should be called and when they shouldn't be.
       - Explain what to do, and not do, if a user has invoked the skill in Copilot outside of Excel.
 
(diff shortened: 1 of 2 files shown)


docs/excel/excel-skills.md
```diff
@@ -0,0 +1,159 @@
...
+
+The main guidance for creating skills is at [Build plugins for Copilot Cowork](/microsoft-365/copilot/cowork/cowork-plugin-development). The context of the article refers to Microsoft Cowork, but the guidance applies generally to all custom skills that can be plugged into an agent running in Copilot.
+
+Start with the section [What you'll build](/microsoft-365/copilot/cowork/cowork-plugin-development#what-youll-build) and then skip to [Build a plugin from scratch](/microsoft-365/copilot/cowork/cowork-plugin-development#build-a-plugin-from-scratch). 
+
+*This article provides only supplementary guidance for creating skills that use Office.js to interact with an Excel workbook.*
+
+## Supplement to Step 1
+
+1. In the YAML frontmatter of a SKILL.md file, include a `metadata.tags` property with an `excel` tag, and other tags as needed. The following is an example.
```
samantharamon:
```suggestion
1. In the YAML frontmatter of a `SKILL.md` file, include a `metadata.tags` property with an `excel` tag, and other tags as needed. The following is an example.
```

For formatting consistency with the other article.

docs/excel/excel-skills.md
```diff
@@ -0,0 +1,159 @@
...
+You can create custom skills that use the Office JavaScript Library (Office.js) to accomplish complex tasks for AI agents.
+
+The main guidance for creating skills is at [Build plugins for Copilot Cowork](/microsoft-365/copilot/cowork/cowork-plugin-development). The context of the article refers to Microsoft Cowork, but the guidance applies generally to all custom skills that can be plugged into an agent running in Copilot.
+
+Start with the section [What you'll build](/microsoft-365/copilot/cowork/cowork-plugin-development#what-youll-build) and then skip to [Build a plugin from scratch](/microsoft-365/copilot/cowork/cowork-plugin-development#build-a-plugin-from-scratch). 
+
+*This article provides only supplementary guidance for creating skills that use Office.js to interact with an Excel workbook.*
+
+## Supplement to Step 1
+
```
samantharamon, line 24:
```suggestion

Follow these steps to supplement [Step 1: Create your first skill](/microsoft-365/copilot/cowork/cowork-plugin-development#build-a-plugin-from-scratch#step-1-create-your-first-skill).

```
Clarifies what Step 1 refers to. If you accept this change, please apply the same change to the succeeding sections.

docs/excel/excel-skills.md
```diff
@@ -0,0 +1,159 @@
...
+   ```yaml
+   name: my-excel-skill
+   description: |
+     Use this skill in Excel to ... 
+   metadata:
+     version: 1.0.0
+     tags: excel
+   ```
+
+1. For most Excel skills, the output should be in the workbook, not a chat or CLI interface. So, in place of the "Output Format" section of a SKILL.md file, have a "Workbook Output" section. The following is an example.
```
samantharamon:
```suggestion
1. For most Excel skills, the output should be in the workbook, not a chat or CLI interface. So, in place of the "Output Format" section of a `SKILL.md` file, have a "Workbook Output" section. The following is an example.
```

docs/excel/excel-skills.md
```diff
@@ -0,0 +1,159 @@
...
+   ```md
+   ## Workbook output
+
+   Produce a dashboard sheet with the following characteristics:
+
+   - It should be named "My Skill Output".
+   - ... other characteristics here.
+   ```
+
+1. Include a "Common pitfalls to avoid" section in the SKILL.md. The following is an example.
```
samantharamon:
```suggestion
1. Include a "Common pitfalls to avoid" section in the `SKILL.md`. The following is an example.
```

docs/excel/excel-skills.md
```diff
@@ -0,0 +1,159 @@
...
+   await Excel.run(async (context) => {
+      ...
+      await context.sync();
+   });
+   ```
+
+   > [!IMPORTANT]
+   > The script should *not* call `Office.onReady` or define an `Office.initialize`. Copilot in Excel automatically creates a runtime and initializes Office.js. 
+
+1. Include as resources, two markdown files that will impose some boundaries on how the skill interacts with the workbook.
```
samantharamon:
```suggestion
1. Include two markdown files as resources that will impose some boundaries on how the skill interacts with the workbook. 
```

docs/excel/excel-skills.md
```diff
@@ -0,0 +1,159 @@
...
+
+      - Confirm the selected school, season range, metric choice, and other custom hooks before generating output.
+      - Keep caveats visible for limited games, freshman samples, or incomplete rows.
+      - Label synthetic rows clearly when they are used.
+
+      ```
+   
+   1. The second resource should do the following.
+
+      - Provide some general quality rules that Copilot should always following when using the skill.
```
samantharamon:
```suggestion
      - Provide some general quality rules that Copilot should always follow when using the skill.
```

docs/excel/excel-skills.md
```diff
@@ -0,0 +1,159 @@
...
+      - Use these scripts after the workbook context is ready for shape and sheet generation, not during an analysis-only pass.
+      - If the current environment is outside Excel, explain the workbook steps and do not attempt to execute the Office.js scripts.
+      ```
+
+## Supplement to Step 4
+
+1. Although your skill calls APIs from Office.js, you don't need an `"extensions"` section in the manifest. Copilot in Excel automatically creates a runtime, loads Office.js and executes the scripts in your skill. 
+
+   > [!NOTE]
+   > Because the runtime isn't explicitly configured in an "extensions" element, there is no `"extensions.requirements.capabilities"` key in the manifest, so there is no way to limit the runtime to a version of Excel that supports a specific Requirement Set. It is unlikely, but theoretically possible, that a user's channel may have a version of Excel that doesn't support an API that is called in a script in the skill. The effect of such a call is undefined.
```
samantharamon:
```suggestion
   > Because the runtime isn't explicitly configured in an "extensions" element, there is no `"extensions.requirements.capabilities"` key in the manifest, so there is no way to limit the runtime to a version of Excel that supports a specific requirement set. It's unlikely, but theoretically possible, that a user's channel may have a version of Excel that doesn't support an API that is called in a script in the skill. The effect of such a call is undefined. 
```

docs/excel/excel-skills.md
```diff
@@ -0,0 +1,159 @@
...
+
+   > [!NOTE]
+   > Because the runtime isn't explicitly configured in an "extensions" element, there is no `"extensions.requirements.capabilities"` key in the manifest, so there is no way to limit the runtime to a version of Excel that supports a specific Requirement Set. It is unlikely, but theoretically possible, that a user's channel may have a version of Excel that doesn't support an API that is called in a script in the skill. The effect of such a call is undefined. 
+
+1. Because a skill plugin is an App for Microsoft 365 and is implemented with the unified manifest and package for Microsoft 365, you can include your custom skill in the same manifest and package as any other type of app for Microsoft 365 such as a Teams Tab or an Office Add-in. However, you can't share your Office.js code between the skill and one of these other apps because the skills scripts are in the app package, not hosted online. If you will need to duplicate code in two or more files if it needs to be run by both the skill and the other app.
```
samantharamon:
```suggestion
1. Because a skill plugin is an App for Microsoft 365 and is implemented with the unified manifest and package for Microsoft 365, you can include your custom skill in the same manifest and package as any other type of app for Microsoft 365, such as a Teams Tab or an Office Add-in. However, you can't share your Office.js code between the skill and one of these other apps because the skills scripts are in the app package, not hosted online. You need to duplicate code in two or more files if it needs to be run by both the skill and the other app.
```

docs/excel/excel-skills.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Supplement to Step 7
+
+It is important to cleanly uninstall the skill after each test session. Follow these steps.
```
samantharamon:
```suggestion
It's important to cleanly uninstall the skill after each test session. Follow these steps.
```

docs/excel/excel-skills.md
```diff
@@ -0,0 +1,159 @@
...
+
+1. Open Teams and be sure you're signed in with the same credentials you used to install the skill. 
+1. Select the apps button: the plus sign in a box.
+1. On the **Apps** pane, select **Manage your apps**.
+1. Find your add-in in the list of apps. It has the name specified in the **name** property of the YAML frontmatter in the SKILL.md file.
```
samantharamon:
```suggestion
1. Find your add-in in the list of apps. It has the name specified in the **name** property of the YAML frontmatter in the `SKILL.md` file.
```

samantharamon (approved):
Approving with some suggestions and typo fixes.

docs/excel/excel-skills.md
Rick-Kirkham: Good idea.

mattgeim (approved):

docs/excel/excel-skills.md
Rick-Kirkham: fixed a slightly different way

PR #5821, opened 2026-07-07
Documentation Updates related to Range.values

chandpravandan:
In this PR, we are adding updates to the documentation related to the Range.values API for large ranges

### DIFF
--- a/docs/concepts/resource-limits-and-performance-optimization.md
+++ b/docs/concepts/resource-limits-and-performance-optimization.md
@@ -62,7 +62,7 @@ Outlook add-ins that use regular expressions and run in Outlook on Windows (clas
 Excel add-ins have important data transfer limits when interacting with the workbook.
 
 - Excel on the web has a payload size limit for requests and responses of **5MB**. `RichAPI.Error` will be thrown if that limit is exceeded.
-- A range is limited to **5,000,000** cells for read operations.
+- A range is limited to **5,000,000** cells for read operations. When a read exceeds this limit, the API might return `null` instead of throwing an error. For more information, see [Payload size limit best practices](../excel/performance.md#payload-size-limit-best-practices).
 
 If you expect user input will exceed these limits, check the data before calling `context.sync()`. Split the operation into smaller pieces as needed. Call `context.sync()` for each sub-operation to avoid those operations getting batched together again.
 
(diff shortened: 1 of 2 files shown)


docs/excel/performance.md
```diff
@@ -126,6 +126,9 @@ await Excel.run(async (context) => {
 
 The Excel JavaScript API has size limitations for API calls. **Excel on the web** limits requests and responses to **5 MB**. The API returns a `RichAPI.Error` error if this limit is exceeded. On all platforms, a range is limited to five million cells for get operations. Large ranges often exceed both limits.
 
+> [!NOTE]
+> When a get operation using `Range.values` exceeds the five-million-cell limit, the API doesn't always throw an error. In some cases, `Range.values` returns `null` instead. To avoid this, check the address of the range before you read `Range.values`. If the range exceeds the limit, read it in multiple smaller batches.
```
AlexJerabek, line 130: What is the actual reason null might be returned instead of an error? I'd prefer to avoid ambiguity in the docs, if possible.

chandpravandan: sorry for the late reply, actually it would be dependent on the range also, like if the size limit of the data is being increased or the number of cells, so from a developer point of view it would be safer to check for both these return values

docs/concepts/resource-limits-and-performance-optimization.md
```diff
@@ -62,7 +62,7 @@ Outlook add-ins that use regular expressions and run in Outlook on Windows (clas
 Excel add-ins have important data transfer limits when interacting with the workbook.
 
 - Excel on the web has a payload size limit for requests and responses of **5MB**. `RichAPI.Error` will be thrown if that limit is exceeded.
-- A range is limited to **5,000,000** cells for read operations.
+- A range is limited to **5,000,000** cells for read operations. When a read exceeds this limit, the API might return `null` instead of throwing an error. For more information, see [Payload size limit best practices](../excel/performance.md#payload-size-limit-best-practices).
```
AlexJerabek:
```suggestion
- A range is limited to **5,000,000** cells for read operations. When a read exceeds this limit, the API will either return `null` or throw an error. For more information, see [Payload size limit best practices](../excel/performance.md#payload-size-limit-best-practices).
```

AlexJerabek (approved):

PR #5839, opened 2026-07-21
docs: Fix sample link in unified manifest feedback doc

MarianKijewski:
Updated the link for the sample App for Microsoft 365 that combines a Teams app and Outlook add-in.

The resource appears to be archived. If you plan to reject this, then remove the link rather than keeping the broken one.


Rick-Kirkham (approved):
I will approve to fix the link. I'm investigating why the sample was archived.

